Issue #90, opened 2013-04-02
Client object type varies based on symfony environment


jkm9000:
Not sure if this is expected behavior, something I am doing wrong, or what... I have a service where I inject redis and when it is in dev mode the object is of type \Snc\RedisBundle\Client\Phpredis\Client but in prod mode it is of type \Redis. So I can't use type hinting in the service constructor.

My configs are below:

```
    snc_redis:
        clients:
            cache:
                type: phpredis
                alias: cache
                dsn: redis://localhost
                logging: %kernel.debug%
        session:
            client: cache
            prefix: sessions
            use_as_default: true

acme.teams_recent:
    class: Acme\TeamBundle\Service\RecentTeams
    arguments: [@snc_redis.cache]
```

fabienwarniez:
Any plans to release this? Right now it's only in master.

ibrahimab:
I would also like to know when this is going to be released?

PR #126, opened 2014-01-30
Restructure the phpredis client.

antoox:
**Fix the issue: https://github.com/snc/SncRedisBundle/issues/90**

On the Phpredis Client directory:
We have now a **ClientInterface**, a **BaseClient** class (simple proxy of the \Redis object) and a **Client** class (extends _BaseClient_) which works as before.

When logging is set to _false_, we instantiate a _\Snc\RedisBundle\Client\BaseClient_ otherwise it is a _\Snc\RedisBundle\Client\Client_. On the application, we can type hint for _\Snc\RedisBundle\Client\ClientInterface_ (and to be sure that logs are activated for this object we can check the type of object _(\Snc\RedisBundle\Client\PhpredisClient)_).

A new option **"phpredis_base_connection_wrapper"** is available and allows to use your own base class client. 

This implementation is still open on modifications. Any comments would be grateful.

Cheers,
Yassine Khial


curry684:
Looking at the code I'm expecting it was removed before the release of PHP7 as it broke this approach due to how it more strictly evaluates function signatures in inheritance.

PR #127, opened 2014-02-17
Client node not required by default

goetas:
Hi!
Sometimes (especially working with different 'environments') is useful disable redis bundle (via config.yml) but currently not possible.


goetas:
any news?

Seldaek:
Merged with an additional fix so it doesn't require at least one client, that way you can configure it as `clients: ~`. I find this is quite useful when doing overrides in parameters, you can have `clients: %redis_clients%` and then define in parameters based on the environment.

Issue #119, opened 2013-12-20
Create command to clear Doctrine cache only


TNAJanssen:
When we have an application running in production mode, we cache a lot through Redis including sessions. When we clear the cache because of a entity update (adding a field or renaming a field). We have the problem that if we don't clear Redis the new field isn't found by doctrine, and if we do clear the cache all sessions are gone.

thvd:
:+1:

AlmogBaku:
`bin/console doctrine:cache:clear-metadata`

JHGitty:
Question answered. Please close this issue. :) #263

JHGitty:
@Seldaek You have reopened this issue. What do you think is missing here? :)

Seldaek:
Nothing, must have pressed twice by accident

PR #129, opened 2014-03-05
Created task to delete redis keys with specified prefix [refs #119]

tomcyr:


tomcyr:
@snc I don't know what to do. I tried git rebase one more time but it looks not good :(

peterjmit:
@tomcyr @snc I don't think this should be merged as is. It is not recommended that the `keys` command be run in production (see [this warning](http://redis.io/commands/KEYS)).

curry684:
I'm closing this PR as it has become unmergeable by now, and should be rewritten to use [`SCAN`](https://redis.io/commands/scan) instead of `KEYS` for the reason mentioned above.

I'm also not sure I see the use case of the command as a whole as it implies highly application-specific applicability, which should usually be implemented in safer ways, which it has been for the original use case mentioned in #119.

PR #141, opened 2014-06-15
persistent connections - use a different persistent_id per DSN

nail:
When using persistent connections with different clients using the same redis server but different databases, the call to 'select' will only be issued during the client's construction, and it will happen on the same persistent connection. You'll end up working on the last opened client's database.

To fix this, I propose to use the persistent_id that can be given to pconnect (https://github.com/nicolasff/phpredis#pconnect-popen) and to tie the persistent connection to the full dsn (using an sha1 of the DSN).

### DIFF
--- a/DependencyInjection/Configuration/RedisDsn.php
+++ b/DependencyInjection/Configuration/RedisDsn.php
@@ -128,6 +128,14 @@ class RedisDsn
     }
 
     /**
+     * @return string
+     */
+    public function getPersistentId()
+    {
+        return sha1($this->dsn);
+    }
+
+    /**
      * @return bool
      */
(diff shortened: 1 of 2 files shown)


pedro-veiga:
Hello, I'd like to ask why was this addition for PhpRedis removed shortly after in commit [Update for PRedis 1.0](https://github.com/snc/SncRedisBundle/commit/6c9d8d7dff1a9d62006bba3c7f0cf395220129a9#diff-7f3f49b43860946c47891a1082cba170) .
Was it removed by mistake perhaps? Or is there a better way to provide a _persistent_id_ for PhpRedis clients? 
Thank you.

PR #152, opened 2014-08-17
use psr-4 autoloader

jrobeson:
since target-dir is deprecated


snc:
Thanks, this was already fixed in #215.

PR #158, opened 2014-09-23
Incorrect prefix using during lock/unlock operation

nightphobos:
Everything is actually working, but in logs we can see confusing LOCK records to "PrefixKey", and SET records to "Prefix:Key", and, if we have getRedisKey function - let's use it.


snc:
Thanks.

PR #160, opened 2014-09-26
doc update, the bundle should be registered last

t0k4rt:
Maybe it should be specified that the bundle should be registered at the end of your appkernel file since we override some services aliases provided in symfony framework bundle.
Actually I spend one hour figuring out why the bundle wouldn't work because of that mistake ^^


TomAdam:
It doesn't need to be registered at the end, just after the framework bundle. Standard practice is to register most bundles after the core SF bundles. I don't think many people will hit this issue.

Issue #159, opened 2014-09-23
DSN with slave instances not working as expected


penfold45:
Hi

I have just implemented using a slave redis insatance and I have the following settings

<pre>
snc_redis:
    clients:
        default:
            type: predis
            alias: default
            dsn:
               - redis://%redis.location%/0?alias=master
               - redis://%redis.slave.location%/0
            options:
                replication: true
            logging: %kernel.debug%

        session:
            type: predis
            alias: session
            dsn:
               - redis://%redis.location%/1?alias=master
               - redis://%redis.slave.location%/1
            options:
                replication: true
        profiler:
            type: predis
            alias: profiler
            dsn:
               - redis://%redis.location%/2?alias=master
               - redis://%redis.slave.location%/2
            options:
                replication: true
</pre>

basically the code is now putting all redis keys into database 2 (the last one) and nothing goes into the other.

It does not appear to be directly linked to the alias as the "gets" on the slave are also also looking at database number 2.

PR #165, opened 2014-12-03
Fix multiple connections with same alias overwriting other connections parameters

nzwalker:
This PR fixes a bug where multiple DSNs/connections with the same alias will have their connection parameters overwritten. This is due to the alias being used as a unique key (e.g ```snc_redis.connection.master_parameters```) which is used when building the container definitions for the various client services. This is especially problematic when using master/slave, since Predis relies on "master" being the connection alias.

The fix is to also use the client name to build a unique key for the connection parameters. People might be relying on this bug (either through using the parameters in the container directly, or not realising the connection parameters are being shared), so could be a BC break. Keen to get feedback whether this solution is reasonable.

This fixes issue https://github.com/snc/SncRedisBundle/issues/159

Example problematic configuration below, which exhibits the "cache" client using the wrong prefix option.

````
snc_redis:
    clients:
        session:
          type: predis
          alias: session
          dsn:
              - redis://127.0.0.1:6379/0?alias=master
              - redis://127.0.0.1:6380/0
          logging: %kernel.debug%
          options:
              replication: true
              prefix: 'session'
        cache:
          type: predis
          alias: cache
          dsn:
              - redis://127.0.0.1:6379/1?alias=master
              - redis://127.0.0.1:6380/1
          options:
              replication: true
              prefix: 'cache'
```

### DIFF
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -177,11 +177,11 @@ class SncRedisExtension extends Extension
         $clientDef = new Definition($container->getParameter('snc_redis.client.class'));
         $clientDef->setScope(ContainerInterface::SCOPE_CONTAINER);
         if (1 === $connectionCount) {
-            $clientDef->addArgument(new Reference(sprintf('snc_redis.connection.%s_parameters', $connectionAliases[0])));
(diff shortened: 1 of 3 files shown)


dominics:
This looks like a good bug fix to me.

@snc is this likely to be merged?

Issue #168, opened 2014-12-20
RedisSessionHandler needs to return true (for HHVM) 


karser:
I encountered with this bug, that's true.
Here is the simmiliar discussion:
https://www.bountysource.com/issues/2853769-custom-sessionhandler-needs-to-return-true-for-hhvm

I use:
"snc/redis-bundle": "1.1.*@dev",
hhvm --version
HipHop VM 3.5.0-dev~nightly.2014.12.19 (rel)
Compiler: heads/master-0-g5c4ac76f5d3b2b4b972fd7db1d2580a4cc35ad50
Repo schema: 46c9b799c227a12b6c161fcbd528a745a927babc
Extension API: 20140829

Sander-Toonen:
+1

I can confirm the issue above and the provided pull request does fix the it for me.

curry684:
We're no longer providing support for HHVM following https://symfony.com/blog/symfony-4-end-of-hhvm-support

PR #164, opened 2014-12-03
Fixes session storage for HHVM (and possibly strict PHP environments)

hardchor:
The [SessionHandlerInterface](http://php.net/manual/en/class.sessionhandlerinterface.php) expects boolean return values for all methods (except read). This must've been an oversight when implementing this interface.

Unfortunately, in HHVM, this produces this (extremely unhelpful) error:

```
Warning: Failed to write session data (user). Please verify that the current setting of session.save_path is correct (unix:///var/run/redis/redis.sock) in ...
```

Quick and easy win!

PR #169, opened 2014-12-20
#168 SessionHandler::write should return true

karser:


Koc:
duplicate of #164

trongdong:
How about this? we still have issue with 1.1 branch @snc

snc:
Fixed in 1.1.10.

PR #183, opened 2015-05-22
Update RedisCache.php

gonzalovilaseca:
'get' returns ``false`` not ``null``.
```
    /**
     * Get the value related to the specified key
     *
     * @param   string  $key
     * @return  string|bool: If key didn't exist, FALSE is returned. Otherwise, the value related to this key is returned.
     * @link    http://redis.io/commands/get
     * @example $redis->get('key');
     */
    public function get( $key ) {}
```

### DIFF
--- a/Doctrine/Cache/RedisCache.php
+++ b/Doctrine/Cache/RedisCache.php
@@ -54,7 +54,7 @@ class RedisCache extends CacheProvider
     {
         $result = $this->_redis->get($id);
 
-        return null === $result ? false : unserialize($result);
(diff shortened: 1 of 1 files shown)


Seldaek:
The thing is unserialize(false) also returns false, so right now this correctly returns false no matter if ->get returned null or false. Predis/Client returns null when there is no value, so we need both to work. I'll fix up the code to be clearer though.

Seldaek:
See 9068c22c2b5fca9d2107edf2ce424850a9706e8f

PR #186, opened 2015-06-30
Pass the path parameter to deduplicate persistent connections

### LINKED ISSUE #185: Same underlying connection resource used for multiple clients, database option ignored with persistent connections
Way back in December 2013 (Predis 0.8.5), I contributed a fix for persistent connections that use different databases on the same IP address and port. See nrk/predis#139 for the details. Specifically, until that version, Predis would do this:

``` php
<?php
$cache_client = new Predis\Client('tcp://127.0.0.1:6379?persistent=1&database=0', []);
$durable_client = new Predis\Client('tcp://127.0.0.1:6379?persistent=1&database=1', []);

$cache_client->ping();    # Connection 1, "SELECT" "0", db0 "PING"
$durable_client->ping();  # Connection 1, "SELECT" "1", db1 "PING"
$cache_client->ping();    # Connection 1, db1 "PING" <<< This should have been on db0
$durable_client->ping();  # Connection 1, db1 "PING"
```

My original analysis of what's going on here is useful as an explanation:

> As you can see, initCmds are only run once, upon a connection being established. Then, because the two clients are connecting to the same IP address and port combination, with persistent connections turned on, they share a single persistent connection. The result is that the third ping() is executed in the wrong database (although this is particularly bad when the command is flushdb instead of ping)

The problem is that this bundle hasn't kept up with this fix, and has no way to safely use persistent connections to the same host/port. Here's an example configuration that demonstrates the problem:

``` yaml
clients:
    default:
        type: predis
        dsn:
            - redis://127.0.0.1/0
        options:
            connection_persistent: true
    another:
        type: predis
        dsn:
            - redis://127.0.0.1/1
        options:
            connection_persistent: true
```

With this configuration, using the 'default' and 'another' client together will result in commands being executed on DBs 0 and 1 inconsistently (depending on which client connected most recently, rather than which client instance you use).

This can be _disastrous_. Keys will suddenly disappear on one client instance when another is used, and commands like `flushdb` can destroy data you thought was safe.

This really comes down to this bundle's choice to use the path of the DSN as the database number (Predis more accurately models it as a connection option). The simplest fix is to also pass the database number as the path to Predis' TCP stream initializer.

dominics:
When using persistent connections, Predis should be provided with a 'path' that matches the database number. This will cause the underlying connection resource to be correctly deduplicated, preventing commands from being run on the wrong database.

Fixes #185

### DIFF
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -144,6 +144,9 @@ class SncRedisExtension extends Extension
                 $connection['scheme'] = 'tcp';
                 $connection['host'] = $dsn->getHost();
                 $connection['port'] = $dsn->getPort();
+                if (null !== $dsn->getDatabase()) {
+                    $connection['path'] = $dsn->getDatabase();
+                }
             }
             if (null !== $dsn->getDatabase()) {
                 $connection['database'] = $dsn->getDatabase();


snc:
Can you provide a link where this is documented?

dominics:
https://github.com/nrk/predis/pull/139#issuecomment-30311604

snc:
Thanks, merged.

PR #187, opened 2015-06-30
Update PHPUnit to stable version, load using Composer

dominics:
Instead of relying on the include path, PHPUnit is better listed in `require-dev`. This allows the version of PHPUnit being used to be specified for developers working on the bundle.

Two small fixes for the test suite are included:
 * Prompt autoloading of the Phpredis client before it is mocked
 * Prevent `RedisCacheTest` failures when re-executing tests on the same Redis instance

### DIFF
--- a/Tests/CommandTestCase.php
+++ b/Tests/CommandTestCase.php
@@ -12,6 +12,7 @@
 namespace Snc\RedisBundle\Tests;
 
 use Symfony\Bundle\FrameworkBundle\Console\Application;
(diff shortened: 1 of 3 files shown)


snc:
Is this PR still valid?

dominics:
Yes, you should still specify the version of PHPUnit you expect to use, rather than rely on the user's environment.

I've rebased.

snc:
OK thanks, looks like travis automatically uses the provided version :+1:

PR #189, opened 2015-07-02
Update readme.md

mauriau:
To fixed PHP Fatal error:  Class 'Predis\Profile\Factory' not found in /vendor/snc/redis-bundle/Snc/RedisBundle/DependencyInjection/SncRedisExtension.php on line 158 we have to update predis's version.
$ composer require predis/predis 0.8.x-dev to $ composer require predis/predis 1.0


xabbuh:
The change proposed in #191 looks more complete.

snc:
Thanks, see #210.

PR #130, opened 2014-03-10
Serialization parameters added. #95

mustafaileri:
I added a parameter for serializer.
Can you please add i to document if PR is ok.

Thanks.

PR #199, opened 2015-09-23
serializer option for redis client

edast:
based on #130 
removed all changes not related to serialization option


yellow1912:
So this PR did not passed the test etc, but I think we are in serious need to control the way serializer is used in redis. In my use case I would like to store simple strings in redis that will be read by non php script so I would like to have the option to set serializer to none.

Webonaute:
can we have some fix for this ?

piotrkochan:
There is 4 PR's for Serializer configuration, I can't believe nobody cares about them.

curry684:
@edast we're picking up some overdue late maintenance work, any chance you can rebase?

curry684:
Closed in favor of #411

PR #200, opened 2015-09-24
Change a few Session Handler vars to protected

urg:
Change a few vars to protected for easier implementation if you want to subclass the SessionHandler


urg:
Sorry, forgot 1 method, will close and add later on

PR #202, opened 2015-10-07
Check if "dsns" is an array...

Kevinrob:
...before passing it to "array_map". If not, with a null value, we have a generic PHP Warning and it's hard to debug and find the cause.
With this change, an exception will be throw after.

### DIFF
--- a/DependencyInjection/Configuration/Configuration.php
+++ b/DependencyInjection/Configuration/Configuration.php
@@ -105,6 +105,10 @@ class Configuration implements ConfigurationInterface
                                 ->end()
                                 ->beforeNormalization()
                                     ->always()->then(function($v) {
+                                        if (!is_array($v)) {
+                                            return $v;
+                                        }
+                                        
                                         return array_map(function($dsn) {
                                             $parsed = new RedisDsn($dsn);
 


DependencyInjection/Configuration/Configuration.php
```diff
@@ -105,6 +105,10 @@ private function addClientsSection(ArrayNodeDefinition $rootNode)
                                 ->end()
                                 ->beforeNormalization()
                                     ->always()->then(function($v) {
+                                        if (!is_array($v)) {
+                                            return $v;
+                                        }
```
xabbuh: Isn't this prevented by the code in line 104?

Kevinrob: No, because null is not a string:
`->ifString()->then(function($v) { return (array) $v; })`

Kevinrob:
Somebody check the PR? @snc ?

snc:
Can you post a config where this error occurs?

Kevinrob:
Of course I can!
```yaml
snc_redis:
    clients:
        default:
            type: predis
            alias: default
            dsns: ~
```

This will produce PHP Warning:
`Warning: array_map(): Argument #2 should be an array`
capture

With this PR, there isn't a warning anymore! The value will be `127.0.0.1` by default.

Kevinrob:
@snc I don't know if you see my previous comment?

Kevinrob:
@snc Is something missing for this PR?

snc:
A test would be a nice addition. But I'm not sure about your proposed solution. You write "The value will be `127.0.0.1` by default.", but this default value then comes from predis or phpredis, right? Maybe it would be better to set the default value in the bundle for consistency.

PR #203, opened 2015-10-08
Prevent exception in destructor.

filakhtov:
Attempting to throw an exception from a destructor (called in the time of script termination) causes a fatal error.
Note from: http://php.net/manual/en/language.oop5.decon.php#language.oop5.decon.destructor

Explicit close call that resets redis instance was provided.

Since PHPRedis can throw an exception on close, following situation happens time to time:
PHP Fatal error:  Uncaught exception 'RedisException' with message 'Redis server went away'

### DIFF
--- a/Client/Phpredis/BaseClient.php
+++ b/Client/Phpredis/BaseClient.php
@@ -31,6 +31,8 @@ class BaseClient implements ClientInterface
     public function setRedis(Redis $redis)
     {
         $this->redis = $redis;
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Seldaek:
Please don't mix fixes with functional changes like making methods chainable.. it makes it impossible to merge one without the other. If you can remove the `return $this;` I think this can be merged.

Seldaek:
Actually I just merged #180 that fixed the same problem, but if you still want to add the explicit close method then please rebase other this can be closed I guess.

stloyd:
I think this can be closed as the class don't contain `__destruct()` anymore.

snc:
Thanks @stloyd.

PR #208, opened 2015-10-22
Make Session Handler more extension friendly

urg:
We need to extend the session handler to run a few checks and this change will allow us to extend without needing to fork/duplicate code.


snc:
Is this still needed or do you already use `2.x-dev`?

urg:
I don't have access to that repo anymore (left the company), so not sure if its still in use.  @c-cooperjones might know?

PR #212, opened 2015-11-09
Remove custom styles icon to work well with new profiler design

stloyd:
Broken:
broken icon

Fixed:
fixed


snc:
Thanks for your PR, but I've just merged PR #215 which contains a fix for this, too.

Issue #223, opened 2015-12-04
support for symfony3


Ma27:
symfony 3 was released some days ago, but I'd really appreciate seeing a new major release supporting symfony3 :)

snc:
This is already in master branch but not yet tagged, would be nice if you could test it.

Ma27:
thank you for your fast answer.
I have to wait until I'll do the major version bump to 3.x as there are other bundles that don't support 3.x. But when I do this, I can have a look at your code in dev-master.

lashae:
@snc is there any schedule for tagging the Symfony3 compatible release?

snc:
Would be nice to get some feedback before tagging.

mustashio:
@snc I would like to help. What can I do? Are there any specific tests you want me to run? Do you want tests on a real-life application or clean installation?

snc:
Any form of practical testing would help :+1:

smedbergsimon:
+1

sunra:
+1
stable version supporting Symfony 3 is needed
dev-master - has problems with Symfony Profiler

snc:
@sunra what about opening a new issue describing the problem in detail or even providing a fix for it in a PR?

lashae:
I'm running Symfony3 on PHP7. So far, I can verify the following:

- dev-master is installable via composer.
- I can get the `snc_redis.default` service on my controller.
- Incrementation via `snc_redis.default` service on a Redis index is successful.
- Doctrine metadata is successfully cached on Redis via `snc_redis.cache`.

sunra:
@snc In fact, everything works well
I had not fully configured project

spdionis:
Any idea when support for symfony 3 will be released?

snc:
I'll try to do it this week.

ssalat:
+1

pdefreitas:
+1

desarrolla2:
I have this from clean instalation of symfony3

  Problem 1
    - Installation request for snc/redis-bundle 1.1.x-dev -> satisfiable by snc/redis-bundle[1.1.x-dev].
    - Conclusion: remove symfony/symfony v3.0.1
    - Conclusion: don't install symfony/symfony v3.0.1
[...]

PR #227, opened 2015-12-28
rename symfony2 to symfony

Ma27:
when tagging the new release with sf3, the symfony2 labels are no longer valid since symfony3 is also supported.
In order to avoid confusion, Symfony2 has been renamed to Symfony

Note: similar stuff happened in other projects like symfony itself: https://github.com/symfony/symfony/commit/0a76b7e9fc2fcc0495312047f83843e041e28cfc

#223

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,12 @@
 
 ## About ##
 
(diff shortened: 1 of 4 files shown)


Ma27:
ping @snc

snc:
Thanks!

PR #229, opened 2016-01-03
Add branch alias for master branch

Seldaek:
That way people can require the upcoming 2.0 as `^2.0@dev` instead of `dev-master` which is a lot more future proof.


snc:
Thanks :+1:

Issue #216, opened 2015-11-16
Doctrine Second Level Cache [SOLVED!]


The-Don-Himself:
As always, I must complement the devs on a fantastic bundle, will never get tired of saying it.

Question : 
If possible, how can we use this bundle to activate and store doctrine second level cache? Tried many configuration guesses but none have worked e.g this link http://www.ekreative.com/blog/three-symfony2-database-optimisation-tips-to-implement-today/468-three-symfony2-database-optimisation-tips-to-implement-today

The-Don-Himself:
Does anyone know whether this bundle is compatible with doctrine second level cache? Activating it with redis results in no cache hit put logs does show cache puts. Switching to memcache or apc works but wouldn't it be great to keep everything in redis:)
Ping @nrk

nrk:
Sorry @The-Don-Himself but I'm not familiar with the combination of this bundle with Doctrine, I'm just the maintaner of one of the underlying libraries (Predis) used by SncRedisBunlde. You should wait for @snc to give you a better insight of how things work in your use case.

The-Don-Himself:
@nrk Apologies for the confusion but thanks for responding, fyi Predis rocks!

PouleR:
I'm also struggling with the second_level_cache, however I managed to get it working. The results are stored in Redis and after they are expired they are retrieved from the database again.

However, I'm using multiple Redis DB's, in this case one specific for Doctrine (And for e.g. sessions in another Redis DB). The problem is that the result cache is written to the default Redis DB and not (in this case DB 11)

How can I configure SncRedis / Doctrine to use a separate DSN for this?

config.yml:
```
orm:
    entity_managers:
        default:
            second_level_cache:
                region_cache_driver:
                    type: redis
                enabled: true

snc_redis:
    clients:
        default:
            type: phpredis
            alias: default
            dsn: redis://localhost
        doctrine_redis:
            type: phpredis
            alias: doctrine_redis
            dsn: redis://localhost/11
    doctrine:
        metadata_cache:
            client: doctrine_redis
            entity_manager: default
            document_manager: default
            namespace: "%kernel.root_dir%"
        result_cache:
            client: doctrine_redis
            entity_manager: default
            namespace: "%kernel.root_dir%"
        query_cache:
            client: doctrine_redis
            entity_manager: default
            namespace: "%kernel.root_dir%"
```

Entity.php
```
* @ORM\Cache(usage="READ_ONLY", region="region_entity")
```

[...]

PR #230, opened 2016-01-07
Add Second level Cache, Update Phpredis Client, Deprecate RedisCache, Fix session locking, Fix Travis

Pouix:
I saw that you were going to release a new version, so here is few things that I worked on:
- Add support for Doctrine Second Level Cache
- Rewrite PhpRedis client so it extends \Redis class
- Deprecate RedisCache class in favor of Doctrine\Common\Cache\RedisCache|PredisCache classes
- Fixed session locking: the lock can only be deleted by the one who put it
- Fixed Travis configuration for HHVM

Happy new year!


composer.json
```diff
@@ -15,10 +15,10 @@
     "require": {
         "php": ">=5.3.3",
         "symfony/framework-bundle": "~2.1|~3.0",
-        "symfony/yaml": "~2.1|~3.0"
+        "symfony/yaml": "~2.1|~3.0",
+        "doctrine/cache": "1.*"
```
snc: This is still optional, isn't it?

Client/Phpredis/Client.php
```diff
@@ -113,4 +106,1296 @@ private function flatten($arguments, array &$list)
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clearLastError()
+    {
+        return $this->call('clearLastError', func_get_args());
+    }
+
+    public function _prefix($value)
```
snc: Should be deleted like the other `_*` functions?!

snc:
Thanks for your work, I added some line comments.

composer.json
Pouix: Indeed, in case we only use client or session parts.

Client/Phpredis/Client.php
Pouix: Missed them.

snc:
In the `RedisSessionHandler` class you added LUA code. Older redis versions do not support LUA. I think it would be OK to add a hint to the documentation mentioning the least required redis version, what do you think?

Seldaek:
This needs to be rebased with latest master as well as it conflicts.

Pouix:
@Seldaek It is rebased.

Seldaek:
Thanks, however it's a bit too big for me to review safely right now, I'll let @snc decide if he wants to bring it in.

snc:
I like the refactoring of the PhpRedis class to extend Redis. The only "problem" is see is that logging will not work for new commands (public methods). I'm not sure if a test could/should validate the overriden methods?

Pouix:
@snc At first, I tried to use Reflection and generate the class, but Phpredis Redis class doesn't declare parameters of its public methods (except for `*scan` methods).
Maybe i'm wrong, but I didn't add a test for the same reason.

This is how I'm seeing it:
1. Manual refactoring of \Redis class from Phpredis (what is done with this PR).
2. Add the proper declarations on Phpredis with a PR (which I can't work on right now).
3. Generate and cache the class.

What do you think?

Resources/doc/index.md
```diff
@@ -4,6 +4,10 @@
 
 This bundle integrates [Predis](https://github.com/nrk/predis) and [phpredis](https://github.com/nicolasff/phpredis) into your Symfony application.
 
+## Prerequisite ##
+
+Session handler requires `Redis >= 2.6`.
```
snc: Could you please add "for LUA scripts" here?

snc:
So when reflection currently can't be used, I'm OK to merge this PR as is.

Resources/doc/index.md
Pouix: This is fixed.

snc:
Merged, big thanks.

Pouix:
You're welcome.

The-Don-Himself:
Is there a specific doctrine configuration for using the second_level_cache. I have been testing this out but keep getting exceptions such as 

````
Catchable Fatal Error: Argument 1 passed to Doctrine\Common\Cache\PredisCache::__construct() must implement interface Predis\ClientInterface, none given
````

snc:
Looks like a bug, see #233.

The-Don-Himself:
Ok, will test again once commit is merged and report accordingly.

The-Don-Himself:
Can confirm that #233 fixes this. Queries are successfully cached and db updates automatically updates the cache as well.
Will have to move it to prod to test replication but think it should work flawlessly. Don't have a cluster step up though.
Funny thing though, I can't seem to see the cache ttl in the logs, doctrine second level has a 3600 default ttl with the previous method in given in #216 I would see this number in the logs but now I don't Will investigate more think its just an issue on my end.

PR #231, opened 2016-01-08
Fix #228 RateLimit and add some tests

### LINKED ISSUE #228: Predis Client::multiExec no longer exists, but is still used in RateLimit
As you can see in the Predis changelog entry for v0.8.5 (2014-01-16):

> Predis\Client::multiExec(): superseded by Predis\Client::transaction() and to be removed in the next major release.

https://github.com/nrk/predis/blob/84060b9034d756b4d79641667d7f9efe1aeb8e04/CHANGELOG.md

However, the `multiExec` method is still used in the `RateLimit` class:

https://github.com/snc/SncRedisBundle/blob/6d513735ec0465013ec51b191e21894f057eb683/Extra/RateLimit.php#L35

https://github.com/snc/SncRedisBundle/search?utf8=✓&q=multiexec

I noticed this because I'm getting this error when trying to use `RateLimit`:

> Command 'MULTIEXEC' is not a registered Redis command. 

I assume this error is related to the non-existent function.

Potentially related to #222.

Pouix:
Fix issue #228 
- Replace Predis\Client::multiExec with Predis\Client::transaction
- Few CS fixes
- Add tests

### DIFF
--- a/Extra/RateLimit.php
+++ b/Extra/RateLimit.php
@@ -12,7 +12,7 @@
 namespace Snc\RedisBundle\Extra;
 
 use Predis\Client;
-use Predis\Transaction\MultiExecContext;
+use Predis\Transaction\MultiExec;
 
 /**
  * This class is a PHP port of the RateLimit structure from the high-level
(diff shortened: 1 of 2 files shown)


snc:
Thanks for this PR, but I think it would be OK too to delete this class. There is https://github.com/jaytaph/RateLimitBundle as a replacement. What do you think?

Seldaek:
Merging because as the code is there already I don't think it's harmful to keep it around and up to date. RateLimitBundle is very symfony-action oriented which IMO isn't necessarily enough depending on what you do and where you want to rate limit. This is a nice simple solution.

PR #232, opened 2016-01-11
Don't test not supported Symfony versions

stloyd:


Pouix:
Good idea.
Could you update composer.json too?

Seldaek:
Thanks, and I fixed the composer.json

PR #233, opened 2016-01-19
fix create doctrine cache difinition when predis is client

shustrik:
Fix doctrine cache definition when cache client is predis. Doctrine\Common\Cache\PredisCache has constructor with ClientInterface parameter, but now client is set from method injection setRedis.


snc:
Symfony 2.7 has PHP `>=5.3.9` as requirement, could you please switch to the old `array()` syntax so that it runs with those old versions?

shustrik:
fixed

snc:
Would be nice if you could squash your commits into one.

shustrik:
fixed

snc:
Nice, thanks.

Issue #235, opened 2016-01-19
strlen() expects parameter 1 to be string, array given in RedisSessionHandler lockSession


qwebot:
I have a problem. My config looks like this:
```yaml
snc_redis:
    clients:
        session:
            type: predis
            alias: session
            logging: false
            dsn: redis://localhost
        cache:
            type: predis
            logging: false
            alias: cache
            dsn: redis://localhost/1
    session:
        client: session
        ttl: 86000
        prefix: foo
        use_as_default: true
```

without description doctrine part. So, in this case there is a warning in lockSession method RedisSessionHandler class. When it call $this->redis->set(..) the third argument is array('NX', 'PX' => $this->lockMaxWait * 1000 + 1) , but it must be a string.

Pouix:
What is your Redis and Predis version ?

qwebot:
Redis 2.2.5
Predis 1.0.3

prodejledu:
I have this problem too. My versions are Redist 2.8.23 and Predis 1.0.3 .

snc:
I've just merged a PR that should fix this, can you validate please?

Pouix:
@qwebot The session handler requires Redis >= 2.6.12

qwebot:
@snc @Pouix  thx! now it works fine!

prodejledu:
@snc works fine, thanks!

PR #236, opened 2016-01-19
Fix Session Handler with Predis

Pouix:
Fix Predis SET function with options (#235).
Set Previs default profile with `default` value.
Fix Prerequisite section in the documentation.

### DIFF
--- a/Session/Storage/Handler/RedisSessionHandler.php
+++ b/Session/Storage/Handler/RedisSessionHandler.php
@@ -119,11 +119,24 @@ class RedisSessionHandler implements \SessionHandlerInterface
         for ($i = 0;$i < $attempts;++$i) {
 
             // We try to aquire the lock
-            $success = $this->redis->set(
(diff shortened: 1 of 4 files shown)


Session/Storage/Handler/RedisSessionHandler.php
```diff
@@ -119,11 +119,24 @@ private function lockSession($sessionId)
         for ($i = 0;$i < $attempts;++$i) {
 
             // We try to aquire the lock
-            $success = $this->redis->set(
-                $this->prefix.$this->lockKey,
-                $this->token,
-                array('NX', 'PX' => $this->lockMaxWait * 1000 + 1)
-            );
+            $setFunction = function (&$redis, $key, $token, $ttl) {
```
snc: No need for the `&` here as it should always be an object.

snc:
Merged, thanks.

PR #238, opened 2016-01-29
Allow easier inherritence of RedisSessionHandler

urg:
We have a custom session handler that inherits from RedisSessionHandler.  Changing these vars/methods to protected will allow us to avoid needing a custom fork.


snc:
OK, merged, thanks.

PR #244, opened 2016-02-09
Remove custom styles icon to work well with new profiler design

stloyd:
Broken:
broken icon

Fixed:
fixed

This is updated version of #212, cause that bug still exists (tested Mac + Chrome).


snc:
Thanks, I'm not sure when `profiler_markup_version` is `1`, do you know this?

stloyd:
When codebase uses Symfony `2.7` not `2.8+` profiler: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php#L115

snc:
Ah OK, thanks.

PR #247, opened 2016-02-22
Rewritten DSN parsing to use parse_url and parse_str to comply with rfc3986 (fixes #218)

### LINKED ISSUE #218: Auth information from DSN read incorrectly
According to RFC 3986, a user info part without a password means that part is the user name, not a password.

Currently, the bundle does not accept `redis://irrelevant:realpass@host` as a DSN, and treats the "foobar" in `redis://foobar@host` as the password, which is in violation of the URI spec.

It should instead discard the "irrelevant" bit in the first example above and use "realpass" as the password.

In general, I recommend replacing the custom parser in `Snc\RedisBundle\DependencyInjection\Configuration\RedisDsn` with a combination of `parse_url()` and `parse_str()` (for the query part).

Also see https://tools.ietf.org/html/rfc3986#section-3.2.1 for more details.

NielsJanssen:
See issue #218 for more info. This change makes sure the URL is parsed properly to just extract the password.

### DIFF
--- a/DependencyInjection/Configuration/RedisDsn.php
+++ b/DependencyInjection/Configuration/RedisDsn.php
@@ -156,6 +156,11 @@ class RedisDsn
         if (false !== $pos = strrpos($dsn, '@')) {
             // parse password
             $this->password = str_replace('\@', '@', substr($dsn, 0, $pos));
+
+            if (strstr($this->password, ':')) {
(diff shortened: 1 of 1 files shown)


NielsJanssen:
After the tests failed it seems that the current implementation is the best solution when dealing with sockets as well, I'll try again with a simple fix to extract the password correctly.

PR #248, opened 2016-02-22
Properly extract password from auth credentials (fixes #218)

### LINKED ISSUE #218: Auth information from DSN read incorrectly

NielsJanssen:
Allows urls like postgres://h:password@host.com:1234 to work. I also attempted to rewrite using parse_url & parse_str as suggested in #218, however due to the implementation of socket location parsing, which parse_url doesn't support, this would lead to too much boilerplate code, making it simpler to just add an easy fix.

### DIFF
--- a/DependencyInjection/Configuration/RedisDsn.php
+++ b/DependencyInjection/Configuration/RedisDsn.php
@@ -156,6 +156,11 @@ class RedisDsn
         if (false !== $pos = strrpos($dsn, '@')) {
             // parse password
             $this->password = str_replace('\@', '@', substr($dsn, 0, $pos));
+
+            if (strstr($this->password, ':')) {
+                list(, $this->password) = explode(':', $this->password);
+            }
+
             $dsn = substr($dsn, $pos + 1);
         }
         $dsn = preg_replace_callback('/\?(weight|alias)=[^&]+.*$/', array($this, 'parseParameters'), $dsn); // parse parameters


JHGitty:
Thanks, please merge this!

snc:
Why would you ever want so specify a user? Redis does not support this?

NielsJanssen:
It is the URL format provided by all of Heroku's Redis providers, so this potentially affects a lot of people. There is also pretty much no downside to adding it.

snc:
Well, currently passwords can contain `:` (there is currently no test that handles this). With this change this won't work anymore. The good thing is that 2.0 is not BC, so it would not be a problem.

JHGitty:
@snc Good reason! So we should close this PR - we need to support ``:``.

NielsJanssen:
I would prefer a way to make it configurable then, as it is still a legitimate use case.

snc:
Absolutely, both should be supported. Do you still have your parse_url code around?

NielsJanssen:
I force pushed over it I'm afraid, so no.

Seldaek:
The way to have `:` in passwords is to use `%3A` really. So I would say this is mergeable as is.

Seldaek:
Well, strictly speaking, only the first : is considered as splitting user and pass I think, so it should be changed to `explode(':', $this->password, 2);` to make sure further `:` stay in the password.

Seldaek:
See https://github.com/snc/SncRedisBundle/commit/c5b91aba0947bccb66722e52a9048cf6ca514891 for additional fixes. I hope you're ok with this @snc, I wrote it down in UPDATE.md at least :)

PR #172, opened 2014-12-31
Predis 1.0 in branch 1.1

theUniC:
Hi,

I've reproduced all the work to integrate Predis 1.0 into the ```1.1``` branch.

PR #150, opened 2014-08-01
Updated Predis to work with 1.*, not 0.8

mattmontgomery:
I'm not sure if you'll have any use for this or be interested in any of the changes, but these are changes I made to get SncRedisBundle working with Predis 1.0 -- take them or leave them. Tests look to be passing (with two failures that were present before the changes I've made.)

PR #266, opened 2016-05-05
Increase the cut-limit for a command

goetas:
This will make logs more readable


snc:
Thanks, merged.

Seldaek:
Why not remove this substr()/toStringArgumentReducer entirely? It's the job of the profiler UI to restrict the data shown if needed IMO, but the logs shouldn't be truncated :/ I didn't find any reason why this was added in https://github.com/snc/SncRedisBundle/pull/172 (cc @theUniC).

theUniC:
In #172 I only continued with a previous work to make this bundle work with PRedis 1.0 in #150.

Seldaek:
OK so it was @mattmontgomery who did it, I missed that. Any clue about the issue here @mattmontgomery? :)

Issue #251, opened 2016-03-08
Symfony 2.8+ profiler bar visual glitch


ABM-Dan:
This is mostly an aesthetic issue, but apparently the profiler for Symfony 3.0 is also having issues, which leads me to think that there must be some sort of deprecation in the method used to integrate with the profiler bar.
![screenshot](https://cloud.githubusercontent.com/assets/16050346/13611383/1dd70c98-e517-11e5-867e-c7e8a1bf9ab9.png)

snc:
This should already be fixed in master.

ABM-Dan:
Could this be backported to 1.1 branch?

JHGitty:
@ABM-Dan I think you like to backport these commits:
https://github.com/snc/SncRedisBundle/commit/e13c491dd0578e54be55bb8e1780dde55da49f81
https://github.com/snc/SncRedisBundle/commit/1ffdfd7945703c594b83952661b6253ffdddc48b

Sander-Toonen:
There was still a small glitch visible when no commands where executed. The above pull request should fix this.

NavyCoat:
I upgraded from 2.5.* to 2.8.12 and found that still occurs.
![profiller glitch](https://cloud.githubusercontent.com/assets/13280538/19301815/aacc4320-9061-11e6-9c46-b412c9e18e7b.jpg)

NavyCoat:
I found that still occurs on 1.1.10 with symfony 2.8.12.
![profiller glitch](https://cloud.githubusercontent.com/assets/13280538/19302022/ac13da4e-9062-11e6-8b78-b16cbdebbd0e.jpg)

curry684:
Haven't seen this myself the past year. If this issue still persists for others please feel free to open a new issue.

PR #275, opened 2016-05-31
Don't show wrapper HTML when no icon is visible in profile bar.

Sander-Toonen:
This fixes a glitch in the profile bar icon for the new style profile bar. Item was always visible, even when no icon and text was shown.

#251


Seldaek:
Thanks

PR #288, opened 2016-08-31
Update the maintenance status badge

emirb:
Since `stillmaintained.com` is for sale, here's a simple replacement for the maintenance badge.


Seldaek:
Thanks

PR #289, opened 2016-09-05
make clients configuration optional

bendavies:
Hi,

I'm using this bundle to setup clients for use with doctrine bundle.
In prod I want to use a redis cache, but in dev vm's, we don't use redis (just simple filesystem).

The config currently *requires* the configuration of a client.
I dev, I don't need/want a client, but am forced to define one, so I have to setup dummy config which isn't used anywhere

This change makes `clients` configuration optional, which simplifies my above scenario.

Thanks


ivoba:
You could also move the sncRedis config to config_prod if you dont need it dev env.

PR #297, opened 2016-10-12
Allow setting serializer for Redis client

yellow1912:
Inspired by commit 199 (snc/SncRedisBundle#199), this commits add support
for setting serializer option for Redis client. There are many cases where
setting a specific Serializer (or none) can be very helpful.

The default serializer is set to default, which allows the Adapter to
automatically detect best Serializer to use.


piotrkochan:
Could someone check this PR?

Seldaek:
I'm not sure I understand the trade-offs between this and #199.. what's the upside/downside?

curry684:
Closed in favor of #411

PR #301, opened 2016-11-03
Reimplementation of the RedisProfilerStorage

GijsL:
Because the built-in support for Redis as profiler datastore was dropped as of Symfony3 I reimplemented it in this bundle. It is basically a copy of the original [RedisProfilerStorage](http://api.symfony.com/2.8/Symfony/Component/HttpKernel/Profiler/RedisProfilerStorage.html) and adjusted to handle the client configuration from this bundle. I also copied the phpunit test files.

Let me know if I need to do more

### DIFF
--- a/DependencyInjection/Configuration/Configuration.php
+++ b/DependencyInjection/Configuration/Configuration.php
@@ -66,6 +66,7 @@ class Configuration implements ConfigurationInterface
         $this->addDoctrineSection($rootNode);
         $this->addMonologSection($rootNode);
         $this->addSwiftMailerSection($rootNode);
+        $this->addProfilerStorageSection($rootNode);
 
         return $treeBuilder;
     }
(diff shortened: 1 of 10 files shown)


GijsL:
@Seldaek @snc I would like to see this PR merged. 
Is this bundle still maintained? Is this feature not wanted? Or is there more work to be done?
Please advice

snc:
LGTM... do you know the reason why this was dropped in Symfony3?

GijsL:
No idea, and cant find any discussion on why it was [removed](https://github.com/symfony/symfony/commit/2925df5fe196e4975d6e87b5b54ac5db4c4950e7). 
Maybe the dependency on a php extension.

Pouix:
> In Symfony versions prior to 3.0, profiles could be stored in files, databases, services like Redis and Memcache, etc. Starting from Symfony 3.0, the only storage mechanism with built-in support is the filesystem.

> You can also create your own profile storage service implementing the ProfilerStorageInterface and overriding the profiler.storage service.

Here's what you can read in [the doc](http://symfony.com/doc/3.0/profiler/storage.html) since 3.0 version.

GijsL:
That documentation does not mention why it was removed. I reimplemented it in the way that was suggested, and as a part of this bundle because we already use it in our projects. 

Would be great if this PR could be merged!

oxman:
Interested too

xavierleune:
Yes, in fact support was dropped "because" without further discussion.
It seems it was a matter of simplicity and I regret that the community has not been involved in this choice.

GijsL:
@snc @Seldaek What are your thoughts on this PR? I believe it adds enough value to be merged, but please tell me if you have other thoughts. I need a stable and maintainable solution and this bundle seems the right place.

snc:
Merged.

oxman:
Nice !

GijsL:
Thanks! 
Im not sure what the procedure is for this, but could you release a new version?

GijsL:
@snc Could you release a new version please?

GijsL:
Thanks!

New version https://github.com/snc/SncRedisBundle/releases/tag/2.0.1

PR #305, opened 2016-12-01
removed defaultValue for replication option to allow cluster option

ivoba:
Predis now has built-in support for redis-cluster:
https://github.com/nrk/predis/blob/v1.1/README.md#cluster

However this options has only effect when the replication option is omitted.
But due to the defaultFalse configuration, the option['replication'] was always present and prevented Predis to enter cluster mode:
https://github.com/nrk/predis/blob/v1.1/src/Client.php#L123

An example config is:

```
snc_redis:
    clients:
      default:
          type: predis
          alias: default
          dsn:
              - 'redis://%cache.redis_host.one%:%cache.redis_port%/%cache.redis_db%'
              - 'redis://%cache.redis_host.two%:%cache.redis_port%/%cache.redis_db%'
          options:
            profile: 3.2
            cluster: "redis"
```

With "redis-custer" or "redis" as cluster option commands like this will automatically be resolved:
MOVED 14966 cache.redis_host.two:6381

This probably needs some more work as replication now can be even more than boolean:
https://github.com/nrk/predis/blob/v1.1/README.md#replication

### DIFF
--- a/DependencyInjection/Configuration/Configuration.php
+++ b/DependencyInjection/Configuration/Configuration.php
@@ -135,7 +135,7 @@ class Configuration implements ConfigurationInterface
                                     ->end()
                                     ->scalarNode('cluster')->defaultNull()->end()
                                     ->scalarNode('prefix')->defaultNull()->end()
-                                    ->booleanNode('replication')->defaultFalse()->end()
+                                    ->booleanNode('replication')->end()
                                 ->end()
                             ->end()
                         ->end()


snc:
Does your setup work with #307?

ivoba:
@snc #307 will not work for `cluster: "redis"` because the options array will still have the key "replication" with null and so the isset function will not let the option pass. 
When `->defaultNull()` would be removed it would probably work.

However as i mentioned in #306 i would prefer not to predefine/limit the options by the bundle configuration at all.

jontorrado:
What about merging this PR @snc? This fixes redis cluster for newer predis versions.

ebarault:
@ivoba @snc is there any progress on this?

Is there currently a working config with predis 1.1.1 that allows `cluster: redis` option to work ?

PR #308, opened 2016-12-06
Use persistent id generated from DSN, if connection_persistent is set to 'true'

Donar23:
If multiple connections are used and the option "connection_persistent" is set to true, the second instance currently uses the connection from the first instance. To prevent that from happening, a persistentId is required.

In this fix I just created a md5 hash from the DSN and used it as persistentId.


snc:
Merged, thanks.

Issue #318, opened 2017-01-30
Doctrine ODM service name is still wrong


usmanzafar:
Hey guys as per the ticket #143  this should have been fixed but turns out the latest version still miss these definitions. I was going through the history of the file and seems like as per this [commit](e461fb819b5b44481dda2a6ada60842922cefdb9) everything in reference to ODM definition reverted back without talking about why these are being reverted. 

I might be doing this wrong but can we revert that branch and merge it again, if not then i can send a quick PR if you guys cool with it ?

snc:
Hi, a PR would be welcome.

usmanzafar:
Aye aye captain, just sent a pr fingers crossed :). ps: I love this bundle this is my goto bundle for almost all projects. proud to be part of. cheers

PR #319, opened 2017-01-30
As per the ticket #318, reverting the odm definition change

usmanzafar:
Updated the definitions of doctrine odm from `doctrine.odm` to `doctrine_mongodb.odm`. Also updated the tests to accommodate this change.


usmanzafar:
Hey @snc - let me know your feedback and if all good can you please tag it - keen to have it sorted soon so that we can refer the new release in our codebase :)

snc:
Thanks, merged.

Issue #323, opened 2017-02-15
PHPRedis RedisClient never throws and error


aurimasniekis:
In `PHPRedis/Client::call` method line where is `logCommand` called has always static argument `$error = false`

https://github.com/snc/SncRedisBundle/blob/master/Client/Phpredis/Client.php#L66

snc:
Waiting for your PR, maybe you could add a test for this, too :-)

snc:
PR merged, thanks.

aurimasniekis:
Fixed the code according comments #325

PR #324, opened 2017-02-15
#323 Implemented error handling for PHPRedis/Client::call

aurimasniekis:


Client/Phpredis/Client.php
```diff
@@ -62,8 +62,13 @@ private function call($name, array $arguments = array())
         $startTime = microtime(true);
         $result = call_user_func_array("parent::$name", $arguments);
         $duration = (microtime(true) - $startTime) * 1000;
+        
+        $error = false;
+        if (false === $result) {
+            $error = call_user_func('parent::getLastError');
```
Koc: why wrap in `call_user_func` instead of just call?

aurimasniekis: Yeah, sorry somehow just copied from above because getLastError is also wrapped

PR #329, opened 2017-02-24
Command executions should be logged as DEBUG instead of INFO. In syst…

FuzzyPones:
…ems where INFO is being used to log informative messages like 'user logged in'.A loop with redis for e.g. in a background-process will pollute the logs extremely.

Commands should be saved as DEBUG logs entries rather than INFO. As INFO is for informative messages and DEBUG for debugging purposes.
A loop with a lot of redis calls (a daemon for example) could easily pollute the logs with command execution messages.
While the next level is WARNING a mesage like "background loop process started" is not WARNING and not DEBUG, i consider that as an INFO log entry.
It would also be consistent with Doctrine for example who log their commands as DEBUG.


Seldaek:
I agree the log level is too high but that probably belongs in a major release as it's kinda breaking.

Seldaek:
However, 876f722 should not be part of this PR IMO

FuzzyPones:
Yeah sorry, i kinda messed up the pull request somehow :(

curry684:
@FuzzyPones can you rebase the PR and remove the extra commit?

FuzzyPones:
Ok should be better now i hope.

curry684:
Manually merged, thanks!

PR #330, opened 2017-02-26
Possible fix for bundle running as daemon

FuzzyPones:
Only collecting the last 100 commands in CLI. This way the bundle won't consume all memory.

### DIFF
--- a/Logger/RedisLogger.php
+++ b/Logger/RedisLogger.php
@@ -50,7 +50,14 @@ class RedisLogger
         ++$this->nbCommands;
 
         if (null !== $this->logger) {
+
+            // if run in CLI only save the last 100 commands to prevent memory leaks in daemon like processes
+            if (php_sapi_name() === 'cli' && count($this->commands) > 100) {
+                array_shift($this->commands);
+            }
+
             $this->commands[] = array('cmd' => $command, 'executionMS' => $duration, 'conn' => $connection, 'error' => $error);
+
             if ($error) {
                 $message = 'Command "' . $command . '" failed (' . $error . ')';
 
--- a/Logger/RedisLogger.php
+++ b/Logger/RedisLogger.php
@@ -62,7 +62,7 @@ class RedisLogger
                     $this->logger->err($message);
                 }
             } else {
-                $this->logger->info('Executing command "' . $command . '"');
+                $this->logger->debug('Executing command "' . $command . '"');
             }
         }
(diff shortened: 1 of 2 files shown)


curry684:
Shouldn't we just disable the logging altogether in CLI?

Seldaek:
I'd say no.. if you run in dev env you get dev env, and I'd argue if you have a logger attached you might be interested in debugging even in a CLI context then we should indeed log things to it. The main problem is that it keeps commands forever on the instance, not that it forwards them to `$this->logger`.

curry684:
Ok, but if we opt for that we should keep in mind that you can also enable logging in prod, and that 100 is a pretty low arbitrary number. I've seen web requests hit 200+ as well. 

So I'd be more in favor of a sufficiently high limit that is still reasonably friendly on memory, like 1000 or even 2500, perhaps configurable with sane high default. Low magic thresholds are dumbfounded devvers waiting to happen.

Seldaek:
No my point was I don't see why this is keeping queries at all on the instance. Maybe it's for the datacollector in dev mode, but in prod mode having the logger enabled should not do collect queries on the instance as that is what leaks memory.

curry684:
Ah k then we mean to say the same thing.

@fuzzypones do you agree and can you change the PR?

FuzzyPones:
Yes i agree.
I am a bit confused as to what to be changed now.
Do we forward it to $this->logger now in DEV, set it to something like 2500?

curry684:
I'm actually not even sure why we should keep `RedisLogger` around if we can just push right to a PSR logger...?

FuzzyPones:
I can only guess since i don't have any numbers but if it is being used in environments that do not have a PSR logger (for any reason) you would have at least simple logging.
I certainly don't know how many usecases there are out there of that.
It certainly was added before any sophisticated logging interfaces were around.
Bundled with the most frameworks it certainly does not make much sense anymore imho.

curry684:
Yes actually the reason it couldn't merge anymore was that I removed the Symfony 2.3 logger, after that it seems to be just a simple PSR forwarder. As PSR is a standardized interface I don't really see a reason anymore we'd need to keep the facade around other than legacy, and we're breaking that anyway in 3.0.

curry684:
Ghe yes well we'll still need to change stuff, but this PR as such may not be the venue anymore indeed  😉

PR #333, opened 2017-03-04
Allow Redis DSN with DB as env param

stingus:
The RedisDSN parser allows only integer values for the DB part in the DSN. In Symfony 3.2, when using an env var for the Redis db parameter, the DNS value is yet not replaced with the actual value taken from env, but a string like `param_name_f8e13c0dd01c2d40bb147771c5d38582`. This prohibits the use of a env var for the DB parameter, like:
`dsn: 'redis://%redis_password%@%redis_host%/%redis_db%'`
This PR fixed this behavior by making a broader match on the DSN.

### DIFF
--- a/DependencyInjection/Configuration/RedisDsn.php
+++ b/DependencyInjection/Configuration/RedisDsn.php
@@ -126,11 +126,11 @@ class RedisDsn
     {
         return $this->alias;
     }
-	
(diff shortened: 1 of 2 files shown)


stingus:
Fixed in 2.0.2

PR #337, opened 2017-04-04
Fixed usage of non-Twig paths

stellalie:
@snc This is a fix so we can use this bundle without twig integration with symfony.

It seems like this template naming is encouraged.

References:
https://github.com/symfony/web-profiler-bundle/commit/50d6db262e70d0e1be39ed914db19d8b3f7a645e


snc:
Thanks 👍

PR #341, opened 2017-05-15
Move the LICENSE file to the root

stof:
This will allow Github to detect the license of the repo, by following the standard convention.

Note that the Symfony best practices for shared bundles have been updated a while ago to recommend putting the license at the root (we changed this after Github announced their support of the license detection)


snc:
Thanks, I've merged it.

PR #345, opened 2017-05-23
adding autodiscover

oleg-andreyev:
Adding autodiscover option

### DIFF
--- a/DependencyInjection/Configuration/Configuration.php
+++ b/DependencyInjection/Configuration/Configuration.php
@@ -136,6 +136,7 @@ class Configuration implements ConfigurationInterface
                                     ->end()
                                     ->scalarNode('cluster')->defaultNull()->end()
                                     ->scalarNode('prefix')->defaultNull()->end()
+                                    ->booleanNode('autodiscovery')->defaultTrue()->end()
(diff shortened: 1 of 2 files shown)


curry684:
Is this compatible with PhpRedis?

oleg-andreyev:
It was long time ago, I'll have to recheck

oleg-andreyev:
Seems that PhpRedis does not have native support of autodiscovery not replication (https://github.com/phpredis/phpredis/issues/663) and should be done on application level

curry684:
Expected as much, but I'm afraid that means this PR would or could break things when used with PhpRedis? At the very least we'd have to issue some sort of warning that this feature is not available, or throw an explicit error perhaps.

curry684:
I'm still open to merging this feature, but it would have to do proper checking that it is only applied to Predis connections.

oleg-andreyev:
@curry684 sorry, I didn't had time to review PhpRedis, I'll try to make it be the end of the week.

curry684:
Last call for changes or additions, closing next time.

oleg-andreyev:
Sorry. I haven't worked with redis for a long time.

PR #349, opened 2017-06-05
Fix template namespacing

alOneh:
I'm using this bundle in a Symfony 3.3 project and I have an issue when using the web profiler due to the wrong template syntax use. The twig loader is not able to find the template.


snc:
Merged, thanks.

Issue #105, opened 2013-10-24
Exception handling if redis is not avaialble


billybigpotatoes:
Can the bundle allow for an exception to be thrown so that Symfony can deal with redis servers not being available gracefully. 

We use redis for caching - so in the event of the back end redis server not being available - we can still generate the content rather than returning a 500 server error

maurotdo:
You can extend the RedisCache, and then register the extended class as a custom provider. 
In the extended class you can do whatever you want, from ignoring the exceptions to inject a failback / callback

jessegreathouse:
@epicwhale, @rdohms, @billybigpotatoes, @snc, Sorry I'm late to the discussion but I was just discussing this issue with one of my engineers who was designing automated health checks meant to alert our support staff when any of our site's components are faulty.

I understand all the points that you have made and what you say is very sensible, and I do not necessarily agree that a "fallback" session storage is appropriate, but let me describe a situation that I feel is untenable in current status quo.

Our site falls into group number 1 that you described, we have multiple nodes and use Redis to have a common session storage between nodes. Indeed, if Redis goes down, we have bigger problems than sessions, but the situation we encountered today gave me a great deal of despair. We created an endpoint on our app which tests for redis usability and in a try catch block where a failure will not throw an exception, but rather give us a json response indicating that the health for the Redis component is critical. The problem that we have is that because symfony is using Redis for sessions, it throws a 500 error regardless.

There needs to be some way that the application can still function when session storage goes away, even if it's only to alert us that session storage has gone away. What are your thoughts?

Garfield-fr:
I have the same problem hier. I use the session storage.  If my redis server is down, i would like to catch this and return Error 500.

What's the best solution for that ? Any implementation in bundle ?

Thank's for your help

maximivanov:
Refreshing this thread as I also don't see a way to gracefully handle redis connection exception. In the project I'm working on, redis is an optional dependency (used to collect stats), and if it's not available site shouldn't crash.
Having this basic setup in config:
```
snc_redis:
    clients:
        default:
            type: phpredis
            alias: default
            dsn: '%redis_dsn%'

AppBundle\Service\Customer\StatsCollector:
    arguments:
        $redis: '@snc_redis.default'
```
and stopping redis docker container, I'm getting `ContextErrorException
Warning: Redis::connect(): php_network_getaddresses: getaddrinfo failed: Name or service not known`.

Looking at exception trace, it happens when redis service is initialized in symfony container.
[...]

PR #362, opened 2017-10-06
Adding lazy configuration for LazyLoading Redis Service - #105

ddproxy:


Seldaek:
I don't see what this is fixing exactly. At least Predis won't connect when you instantiate it so making it lazy saves nothing and adds the overhead of a proxy class.

curry684:
PhpRedis also requires an explicit call to connect to making the services lazy has no added value.

PR #366, opened 2017-10-23
forward compatibility with Symfony 4.0

### LINKED ISSUE #365: Make RedisDataCollector compatible with Symfony 4.0
Symfony will add a `reset()` method to its DataCollectorInterface in order for the profiler data to be reset (see https://github.com/symfony/symfony/pull/24289). Failing to implement this method will trigger a deprecation notice in version 3.4, and break in 4.0 (both versions expected in november).

Example of a 3.4 deprecation notice:

> Implementing "Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface" without the "reset()" method is deprecated since version 3.4 and will be unsupported in 4.0 for class "Snc\RedisBundle\DataCollector\RedisDataCollector"

xabbuh:
This will fix #365.

### DIFF
--- a/DataCollector/RedisDataCollector.php
+++ b/DataCollector/RedisDataCollector.php
@@ -47,6 +47,14 @@ class RedisDataCollector extends DataCollector
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function reset()
+    {
+        $this->data = array();
+    }
+
+    /**
      * Returns an array of collected commands.
      *
      * @return array


snc:
Thanks!

yceruto:
Could we make a new release for this one?

yceruto:
Also `composer.json` should be updated to be able to install symfony/* ^4.0 deps

PR #370, opened 2017-11-08
Provides Symfony 4 compatibility

romain-pierre:
Hi, I'm working to get a compatible version of this bundle for the forthcoming Symfony 4. Don't hesitate to suggest fixes, leave comments about this PR.

PR #373, opened 2017-11-14
Fix ci warning caused by PHPUnit Api deprecation

SiM07:
`getMock()`is deprecated since PHPUnit 5.4.0.

Fix warnings in PR https://github.com/snc/SncRedisBundle/pull/370


snc:
Thanks, merged.

PR #353, opened 2017-06-09
Support for Heroku style REDIS_URL env variables

iKlaus:
Being affected by the issue described in #302 I looked into what needs to be done to support Heroku's REDIS_URL environment variables. In short, they provide the credentials to a booked Redis addon via an environment variable which might change anytime at their discretion:

> In order for Heroku to manage this add-on for you and respond to a variety of operational situations, the REDIS config vars may change at any time. Relying on the config var outside of your Heroku app may result in you having to re-copy the value if it changes.
> -- *https://devcenter.heroku.com/articles/heroku-redis#create-a-new-instance*

Therefor I just wanted to be able to make use of the `env()` dynamic parameter notation as in

```yaml
snc_redis:
    session:
        client: session
    clients:
        session:
            type:  predis
            alias: session
            dsn:  "%env(REDIS_URL)%"
```

I came up with a rather simple approach which basically boils down to:
- detect the specific syntax while the container compiles and bypass the validation on compilation
- assemble the service with the help of a Factory which parsed the url at runtime
- merge parsed url fragments with the rest of the connection options and then instantiates the Predis client as usual

PR #378, opened 2017-12-15
Support env vars for both drivers

### LINKED ISSUE #361: Support for phpredis with RedisEnvDsn is broken
Its missing getSocket, its fixed for predis but not for phpredis!

kozlice:
Resolves #361.

Provides support for `%env(REDIS_URL)%` for both phpredis and PRedis. I'm not sure that resolving variables inside extension is OK, but Twig extension does so. Unfortunately, had to move DSN validation too.

Reverts pretty much everything from #353.


shustrik:
@snc what about this PR?

piotrantosik:
Works perfectly, @snc could you push a new tag with this change?

duraki:
This still not tagged? Any eta? @snc

PR #385, opened 2018-01-18
Protect against session fixation attacks

BboyKeen:
Sometimes, a malicious user will tamper the session cookie to inject a controlled session id.
Without this protection, the session handler tries to read a session value with this session id and, if the id is not found, a new session entry is created using the malicious id.
To avoid that the PHP team introduced the use_strict_mode option in the native sessions module.
This commit introduces the use of the use_strict_mode option in the SessionHandler to prevent session fixation by issuing a new session id for the current session when an unknown session
id is used.
For the sake of facility, we rely only on the opened sessions to conclude whether or not the session id is known.
We could add opened session ids history as in https://github.com/1ma/RedisSessionHandler
but I can't see any useful use case for now.

PR review appreciated

### DIFF
--- a/Session/Storage/Handler/RedisSessionHandler.php
+++ b/Session/Storage/Handler/RedisSessionHandler.php
@@ -73,6 +73,11 @@ class RedisSessionHandler implements \SessionHandlerInterface
     private $lockMaxWait;
 
     /**
+     * @var bool Indicates if the strict mode is enabled
+     */
+    private $useStrictMode;
+
+    /**
      * Redis session storage constructor.
      *
      * @param \Predis\Client|\Redis $redis   Redis database connection
@@ -96,6 +101,7 @@ class RedisSessionHandler implements \SessionHandlerInterface
         if (!$this->lockMaxWait) {
             $this->lockMaxWait = self::DEFAULT_MAX_EXECUTION_TIME;
         }
+        $this->useStrictMode = isset($options['use_strict_mode']) ? (bool) $options['use_strict_mode'] : true;
     }
 
     /**
@@ -197,6 +203,11 @@ LUA;
                     return false;
                 }
             }
+        }
+
+        if ($this->useStrictMode && !$this->redis->exists($this->getRedisKey($sessionId))) {
+            $sessionId = session_create_id();
+            session_id($sessionId);
         }
 
         return $this->redis->get($this->getRedisKey($sessionId)) ?: '';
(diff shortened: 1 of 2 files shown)


Seldaek:
I am wndering if this shouldn't be enabled by default.. Sounds like session fixation isn't something you ever want to allow really.

BboyKeen:
I can't see any case where session fixation would be something wanted, but : 
- we could let it configurable as there is an option in the session configuration in the FrameworkBundle
- we could ensure that session fixation protection is enabled in case no configuration has been set

Sthg like this 
```php
$this->useStrictMode = isset($options['use_strict_mode']) ? (bool) $options['use_strict_mode'] : true;`

BboyKeen:
Well, regarding this thread https://stackoverflow.com/questions/41411367/ensuring-session-strict-mode-in-custom-sessionhandlerinterface-implementation, it looks like the solution proposed is not optimal at all and will introduce BC breaks with older versions of PHP

pascal-hofmann:
I just tested these changes: They do not work. A new session id is generated and used for storage, but PHP still sends the old session ID in headers. So in the end the user gets a fresh session without any data with each request.

I guess the only way to prevent session fixation is to implement `SessionUpdateTimestampHandlerInterface`.

BboyKeen:
I think we have to reissue a session cookie with the new value. I thought the session module of PHP was handling that but it looks like it's not the case.

The problem with the SessionUpdateHandlerInterface is that it's a PHP 7 feature and will introduce BC break

pascal-hofmann:
PHP is responsible for sending the session cookie. I don't think the application should fiddle with it.

PHP 7 is out for more than 2 years now, support for earlier version will end in less 11 months. I think having a BC break is a good thing. People need a reason to upgrade.

And lets not forget: This is a security issue.

Seldaek:
I think the correct way to handle this would be to call SessionStorageInterface::regenerate or SessionInterface::migrate in case the session key does not exist when we load it. Unfortunately there is no way to do this cleanly from within a SessionHandlerInterface implementor in Symfony. Something to be improved on the Symfony side perhaps with a new interface?

BboyKeen:
Started from 3.4, Symfony team provides new session handlers for this issue : 
- https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/AbstractSessionHandler.php
- https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/StrictSessionHandler.php

Seldaek:
Thanks for the pointer @BboyKeen - I implemented that in https://github.com/snc/SncRedisBundle/pull/404 - closing this as it's not the right approach IMO.

BboyKeen:
I perfectly agree. I wi use your PR in production as well.

PR #388, opened 2018-02-02
Updated test with namespaced PHPUnit TestCase

MarioBlazek:
PR makes unit tests extending namespaced `TestCase` rather than `PHPUnit_Framework_TestCase`. This change enables upgrade of PHPUnit to v6, with minimum version of `4.8.35` where compatibility layer was introduced.


MarioBlazek:
@snc please review.

snc:
Thanks!

Issue #390, opened 2018-02-19
Issues with symfony/web-profiler-bundle, SncRedisBundle and content security policy


rpkamp:
When using the bundle in a Symfony project with Content Security Policy configured to reject inline styles/scripts, we get an error because the profiler templates for the symfony web profiler bunle used in snc redis bundle use inline styles.

The inline styles should be removed in order for this bundle to work correctly under CSP.

PR #395, opened 2018-03-02
No longer use style in svg image for profiler

rpkamp:
Using style does not work when a Content Security Policy is in place.
Removing style and just using fill has the same end-result, and also works when a Content Security Policy is in place.

Fixes https://github.com/snc/SncRedisBundle/issues/390


rpkamp:
Build failed due to a Travis error. Closing and re-opening this PR to trigger a new Travis build.

snc:
Thanks, merged 👍

PR #404, opened 2018-03-29
New session handler

Seldaek:
Implement new [Symfony 3.4+ session handler](https://github.com/symfony/symfony/pull/24523) to gain session fixation protection and avoid unnecessary writes.

This breaks compatibility with Symfony 3.3 and below so it should be included in a major version only IMO, but we are going to use it in production soon.

Note that this includes #403 as I didn't want to introduce conflicts but I can easily undo that if needed.

Also note that with this, as Symfony will only write to the session when data actually changed, locking the session becomes a lot less necessary, as you have less chances of overwriting data with outdated one. At least if you don't use the session too much this is a big win in terms of throughput and number of commands redis has to deal with.

So:

```
    session:
        client: session
        ttl: 3600
        locking: false
```

### DIFF
--- a/Session/Storage/Handler/RedisSessionHandler.php
+++ b/Session/Storage/Handler/RedisSessionHandler.php
@@ -11,6 +11,8 @@
 
 namespace Snc\RedisBundle\Session\Storage\Handler;
 
+use Symfony\Component\HttpFoundation\Session\Storage\Handler\AbstractSessionHandler;
+
 /**
  * Redis based session storage with session locking support.
  *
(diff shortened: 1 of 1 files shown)


phansys:
Thank you so much for this feature!
Is there any plan to release a next major including this? Could we have any roadmap for it?
It will bring a great performance impact, so I'm waiting to be able to implement this within a stable release.

Seldaek:
There are a few things left before 3.0 stable can be tagged https://github.com/snc/SncRedisBundle/milestone/2 - it should be relative safe to require 3.0-dev though, we try to keep things stable.

curry684:
Thing is that there's the major breaking change coming up to "fix phpredis logging". Once that one's done it'll mostly imply the other pending changes for 3.0.

We could tag `3.0.0-alpha.1` or something I suppose, at least then you'll be safer than on master branch.

Issue #399, opened 2018-03-19
(bc) problem with 4.0 redis


jroszkiewicz:
PHP Redis 4.0 was released 17-03-2018 and.. we have some bc:

Warning: Declaration of Snc\RedisBundle\Client\Phpredis\Client::append() should be compatible with Redis::append($key, $value)

Proposition:
- release new major version for 4.0 php redis
- in current major version fix composer.json and lock to 3.1.x php redis

curry684:
> lock to 3.1.x php redis

That won't work as the bundle doesn't *depend* on phpredis, since it can also work with predis.

The true solution, which is B/C compatible, is to invoke a different client based on which version is installed.

curry684:
Actually https://ocramius.github.io/ProxyManager/docs/access-interceptor-value-holder.html might do exactly what is wanted here, and it'd be both backwards and future compatible.

curry684:
@snc honest and blunt question as I currently have 4 projects failing CI and going to fail to deploy soon-ish because of this issue: are you still maintaining this project?

I'm happy to join in and get stuff fixed, but not if merging the PRs is going to take months, and given the amount of open issues and PRs, and your Github activity as a whole, I'm getting the impression of an abandoned project. Fine if that is the case, but in that case we have no other choice than to fork.

jroszkiewicz:
@curry684 you can always force to install redis 3.1.x by pecl, buts it's a temporary solution.
However i think symfony/cache is better option. And of course more trusted ;-)

curry684:
`symfony/cache` is a caching solution, not a Redis encapsulation. I have projects using hashmaps, pubsub, ordered lists and everything else Redis offers.

Thanks for the PECL hint, I completely overlooked that option as I didn't expect it to be possible with the stock Docker containers, but it is and it will fix some of my CI issues hehe.

Seldaek:
@curry684 If you send a reasonable PR I'll try and get it in, I have write access here because I remember nagging @snc too once upon a time. I think he's too busy overall :)

curry684:
@Seldaek cheers, I think #405 should be the quick fix. If logging is disabled the wrapper is not used, so I disabled it by default if 4.0.0+ is installed, and issued a warning if explicitly enabled.

curry684:
Well it didn't break tests on my 3.1.6 install, now also fixed tests with 4.0 installed hehe.

chteuchteu:
@Seldaek Could you tag a release please? :-)

[...]

PR #405, opened 2018-03-30
Disable logging wrappers on PhpRedis >= 4.0.0

curry684:
Ref #399


Seldaek:
Not sure why the build is failing.. but it wasn't in master so can you have a look?

curry684:
Funky, it's exactly what we're fixing, but my fix simply omits instantiation of the wrapper on 4.0+, while this tests instantiates it directly hehe.

curry684:
Fixed it.

Issue #407, opened 2018-04-06
[Bug] No commands in profiler using symfony cache with predis


vchebotarev:
Hello.
SncRedisBundle profiler doesn't display command when i'm using symfony cache with snc redis "predis" client only. Going deeper - commands are not even catched.

With "phpredis" everything is fine.

So this is my yml (yes it also possible configuring "framework -> cache -> pools" but in this case it doesn't matter )
```yml
services:
    app.cache.psr6:
        class: Symfony\Component\Cache\Adapter\RedisAdapter
        arguments:
            - '@snc_redis.default'

    app.cache.psr16:
        class: Symfony\Component\Cache\Simple\RedisCache
        arguments:
            - '@snc_redis.default'

snc_redis:
    clients:
        default:
            type:    predis
            alias:   default
            dsn:     'redis://localhost'
            logging: '%kernel.debug%'
```
This is example of usage in controller
```php
        $cachePool = $this->get('app.cache.psr6');
        $cachePoolItem = $cachePool->getItem('app.cache.psr6');
        $cachePoolItem->set('app.cache.psr6');
        $cachePool->save($cachePoolItem);

        $cachePsr16 = $this->get('app.cache.psr16'); //also uses RedisTrait
        $cachePsr16->set('app.cache.psr16', 'app.cache.psr16');

        $sncClient = $this->get('snc_redis.default');
        $sncClient->set('snc_redis.default', 'snc_redis.default');
```

But the profiler looks like this
![e032e7dea3](https://user-images.githubusercontent.com/2445241/38422274-cd094536-39b2-11e8-81c8-6395adb8d721.png)

**Values present in redis. I CHEKED !**

When i try to change client type to "phpredis" everything goes fine
```yml
snc_redis:
    clients:
        default:
            type:    phpredis
            alias:   default
            dsn:     'redis://localhost'
            logging: '%kernel.debug%'
```
![a974b75e04](https://user-images.githubusercontent.com/2445241/38422422-66ca9724-39b3-11e8-9f88-6441c8973516.png)

My application where i demonstrated it is: 
blank symfony 3.4.7
predis/predis v1.1.1
snc/redis-bundle 2.0.6

curry684:
Should be closed by your PR, thanks!

PR #408, opened 2018-04-06
fix predis logging using symfony cache component

vchebotarev:
#407 

Well, i went deeper and found the problem
bdc424d677_ready

It seems that duration is not correct because it doesn't wait for response for the pipeline (as symfony adapters do) ... maybe maiterners will inlarge logger to support it ... 

PS
I've created PR branch from tag 2.0.6 not master

### DIFF
--- a/Client/Predis/Connection/ConnectionWrapper.php
+++ b/Client/Predis/Connection/ConnectionWrapper.php
@@ -95,7 +95,16 @@ class ConnectionWrapper implements NodeConnectionInterface
      */
     public function writeRequest(CommandInterface $command)
     {
-        return $this->connection->writeRequest($command);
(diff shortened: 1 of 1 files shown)


curry684:
Changed base to 2.1 but we can merge back into master.

vchebotarev:
Is there anything I have to do next ?

curry684:
No I'm just hesitant to push changes like this as I'm not too familiar with the Predis integrations (always used phpredis myself). So I'd like @Seldaek or @snc to four-eye quickly before I merge.

Seldaek:
Seems fine to me on a quick look.

curry684:
Then let's try it out   😉

PR #409, opened 2018-04-09
Make all bundle services private

### LINKED ISSUE #384: Alias for services created via dependency injection are not public
First of all: Awesome bundle :)

When an alias is created here:

https://github.com/snc/SncRedisBundle/blob/2.0.6/DependencyInjection/SncRedisExtension.php#L206

it is created in a way that makes it private (it is private by default even if the service is public). By creating it like this

`
        $container->setAlias(sprintf('snc_redis.%s_client', $client['alias']), new Alias(sprintf('snc_redis.%s', $client['alias']), true));
`

it is possible to create it and make it public. 

Are the aliases supposed to be private or do you want a PR? :)

curry684:
Since Symfony 3.4 recommended practice is to keep all services private
and use declarative dependency injection.

Closes #384

### DIFF
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -201,7 +201,7 @@ class SncRedisExtension extends Extension
         $optionDef->addArgument($client['options']);
         $container->setDefinition($optionId, $optionDef);
         $clientDef = new Definition($container->getParameter('snc_redis.client.class'));
-        $clientDef->setPublic(true);
+        $clientDef->setPublic(false);
         $clientDef->addTag('snc_redis.client', array('alias' => $client['alias']));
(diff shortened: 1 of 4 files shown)


umpirsky:
Will https://github.com/snc/SncRedisBundle/blob/master/Command/RedisBaseCommand.php#L67 work after this?

curry684:
Nope, good one, we'll have to fix that.

umpirsky:
I'd love to see how, I have a similar problem in my project. Reported https://github.com/snc/SncRedisBundle/issues/418. Thanks!

curry684:
Symfony offers the concept of a ServiceLocator where you can inject other services to be created lazily   😉

umpirsky:
Great, waiting for the fix!

PR #411, opened 2018-04-12
Add Serialization option

yellow1912:
Serialization option is necessary for setting the type of serialization
to use for redis.

This is simply a new commit with all the code taken from PR#199 using
the latest code of SncRedisBundle on branch 2.1

### DIFF
--- a/DependencyInjection/Configuration/Configuration.php
+++ b/DependencyInjection/Configuration/Configuration.php
@@ -114,6 +114,7 @@ class Configuration implements ConfigurationInterface
                                     ->scalarNode('read_write_timeout')->defaultNull()->end()
                                     ->booleanNode('iterable_multibulk')->defaultFalse()->end()
                                     ->booleanNode('throw_errors')->defaultTrue()->end()
+                                    ->scalarNode('serialization')->defaultValue("default")->end()
                                     ->scalarNode('profile')->defaultValue('default')
                                     ->end()
                                     ->scalarNode('cluster')->defaultNull()->end()
--- a/DependencyInjection/Configuration/Configuration.php
+++ b/DependencyInjection/Configuration/Configuration.php
@@ -114,7 +114,7 @@ class Configuration implements ConfigurationInterface
                                     ->scalarNode('read_write_timeout')->defaultNull()->end()
                                     ->booleanNode('iterable_multibulk')->defaultFalse()->end()
                                     ->booleanNode('throw_errors')->defaultTrue()->end()
-                                    ->scalarNode('serialization')->defaultValue("default")->end()
+                                    ->scalarNode('serialization')->defaultValue('default')->end()
                                     ->scalarNode('profile')->defaultValue('default')
                                     ->end()
                                     ->scalarNode('cluster')->defaultNull()->end()
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -442,11 +442,10 @@ class SncRedisExtension extends Extension
      */
     public function loadSerializationType($type)
     {
-        // not using constants, because it would fail if phpredis extension is not enabled
         $types = array(
-            'none' => 0, // \Redis::SERIALIZER_NONE,
-            'php' => 1,  // \Redis::SERIALIZER_PHP,
-            'igbinary' => 2 // \Redis::SERIALIZER_IGBINARY
+            'none' => \Redis::SERIALIZER_NONE,
+            'php' => \Redis::SERIALIZER_PHP,
+            'igbinary' => \Redis::SERIALIZER_IGBINARY
         );
 
         // allow user to pass in default serialization in which case we should automatically decide for them
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -444,17 +444,13 @@ class SncRedisExtension extends Extension
     {
         // not using constants, because it would fail if phpredis extension is not enabled
         $types = array(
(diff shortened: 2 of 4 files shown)


curry684:
To save us some reviewing effort (see https://github.com/snc/SncRedisBundle/pull/297#issuecomment-377315792 as well), could you explain briefly the improvements over #297 and why you based on #199 instead?

DependencyInjection/Configuration/Configuration.php
```diff
@@ -114,6 +114,7 @@ private function addClientsSection(ArrayNodeDefinition $rootNode)
                                     ->scalarNode('read_write_timeout')->defaultNull()->end()
                                     ->booleanNode('iterable_multibulk')->defaultFalse()->end()
                                     ->booleanNode('throw_errors')->defaultTrue()->end()
+                                    ->scalarNode('serialization')->defaultValue("default")->end()
```
curry684: String constants should be in single quotes unless using interpolation.

DependencyInjection/SncRedisExtension.php
```diff
@@ -427,7 +433,43 @@ protected function loadSwiftMailer(array $config, ContainerBuilder $container)
...
+     * @throws InvalidConfigurationException
+     */
+    public function loadSerializationType($type)
+    {
+        // not using constants, because it would fail if phpredis extension is not enabled
+        $types = array(
+            "none" => 0, // \Redis::SERIALIZER_NONE,
+            "php" => 1,  // \Redis::SERIALIZER_PHP,
+            "igbinary" => 2 // \Redis::SERIALIZER_IGBINARY
+        );
```
curry684: Given the dependency on these constants, why can it even be invoked if phpredis is not loaded?

Also single quotes   ;)

DependencyInjection/SncRedisExtension.php
```diff
@@ -427,7 +433,43 @@ protected function loadSwiftMailer(array $config, ContainerBuilder $container)
...
+
+        // allow user to pass in default serialization in which case we should automatically decide for them
+        if ('default' == $type) {
+            if (defined('HHVM_VERSION')) {
```
curry684: We dropped HHVM support so can drop this even if basing on 2.1 as far as I'm concerned.

Tests/DependencyInjection/SncRedisExtensionTest.php
```diff
@@ -308,6 +308,41 @@ public function testClientReplicationOption()
...
+    public function testLoadSerializationType()
+    {
+        $extension = new SncRedisExtension();
+        $config = $this->parseYaml($this->getSerializationYamlConfig());
+        $extension->load(array($config), $container = $this->getContainer());
+        $options = $container->getDefinition('snc_redis.client.default_options')->getArgument(0);
+        $serializationType = $extension->loadSerializationType($options['serialization']);
+        $this->assertTrue(is_integer($serializationType));
+
+        if (defined('HHVM_VERSION')) {
```
curry684: Same here - it's a new feature so don't need to support HHVM for BC.

DependencyInjection/SncRedisExtension.php
yellow1912: Do you mean to say that we should use  this instead?

$types = array(
            'none' => 0, \Redis::SERIALIZER_NONE,
            'php' => 1,  \Redis::SERIALIZER_PHP,
            'igbinary' => 2 \Redis::SERIALIZER_IGBINARY
        );

DependencyInjection/SncRedisExtension.php
yellow1912: understood, I can remove hvvm part

Tests/DependencyInjection/SncRedisExtensionTest.php
yellow1912: will remove now

yellow1912:
Let me close #297, the reason is that it was from an older version, and since then I made some more changes that shouldn't be in the PR. 

Regarding the other changes you commented I will update now.

Regarding #199, it's minor:
1. Instead of loading all types then check if we support the specific type with loadSerializationTypes, I use loadSerializationType to detect best serialization type to use
2. The default serialization is not none, if no serialization specified it will use try to detect. This will make this change to work nicely with current applications, otherwise you will have to force clean cache after upgrade because the default was not none I believe.

DependencyInjection/SncRedisExtension.php
curry684: The comment says we're using the integers instead of the constants as they would fail if phpredis is not loaded, but in that case the serialization wouldn't work anyway, so how come we have to account for it?

yellow1912:
@curry684 I made all the changes, the only thing left is your comment regarding "Given the dependency on these constants, why can it even be invoked if phpredis is not loaded?". I was unsure what you meant.

yellow1912:
I have updated the code to use Redis class constants directly as per your request.

curry684:
Tests failing on:
```
1) Snc\RedisBundle\Tests\DependencyInjection\SncRedisExtensionTest::testLoadSerializationType
Error: Undefined class constant 'SERIALIZER_IGBINARY'
```
It's apparently optional: https://github.com/phpredis/phpredis/issues/588

Can you fix it so the code is compatible with IGBINARY support missing as it apparently is on Travis?

yellow1912:
Perhaps that was the reason why the hardcoded integer was used instead of the constant, but I think I can fix it.

curry684:
There are [cleaner ways](http://php.net/manual/en/function.constant.php) than hardcoding external integers out of our control   ;)

curry684:
Merged, thanks!

Issue #400, opened 2018-03-22
Connections not properly closed on Kernel shutdown


applike-ss:
I experience propblems because of too many open connections during unit tests.
These connection are at half from ongr which doesn't close the connection on kernel shutdown => Garbage collection doesn't free it.

The used PHP Version is 7.1.12
Bundle is 2.0.6

curry684:
Could you elaborate a bit more on the specific issue you are facing?

curry684:
If I understand correctly => if you have a ton of functional tests instantiating rebooted kernels, and therefore reinstantiating the Redis instances, they will start failing in the end indicating Redis has run out of sockets or something?

curry684:
@applike-ss I've just typed a bit in a related report at https://github.com/snc/SncRedisBundle/pull/412#issuecomment-380919445. As said there I'm not all too excited about changing or extending perfectly fine code just to facilitate test suites if it is easier and more correct to fix the test suite instead.

What is your take on that?

applike-ss:
@curry684 i do understand your point, but it is not the test case/suite that creates the problem. 
In my Opinion, whenever a Bundle is shut down, all connections that this bundle initiated should be properly closed instead of relying on php to close them. 
Doctrine is a good example for this behaviour. When it shuts down, it gathers all entity managers and clears their state. Afterwards it gathers all connections and closes them.
I believe, it shouldn't break anything anyway, since the connection wouldn't be used after a bundle shutdown anyway.

applike-ss:
You are right, having lots of tests using redis, we are facing an issue with connection count increasing to the maximum of the machine running the tests.

supersmile2009:
@applike-ss check out #415 - should be a solution you're looking for.

applike-ss:
@supersmile2009 thanks a lot. Looking forward for it to be merged.

curry684:
Fixed in 2.x by #415 but keeping this issue open to better investigate how to do this in 3.0 without making a service needlessly public.

supersmile2009:
Like @dkarlovi mentioned on #415 it is possible to create env-specific service for test environment and declare it public instead of making public the original service.
However I'm not sure if there's anything else except unit tests that's using Kernel's shutdown feature. Method's docBlock says "This method is mainly useful when doing functional testing." But I'm particularly concerned about things like PPM - I haven't had a chance to try it yet, so I don't know whether it uses shutdown.

[...]

PR #412, opened 2018-04-12
Use destructor to close session (partially reverted #348)

supersmile2009:
Registering a shutdown function to close session in PR #348 was a great idea... However removing a destructor was a terrible one.

We have a pretty heavy test suite in our app and after updating to SncRedisBundle v2.1 (which included that PR) memory usage in it just skyrocketed, PHP started crashing running out of memory.

This PR adds destructor back while also keeping a shutdown function introduced in #348. It solves the memory issue we're facing.

### DIFF
--- a/Session/Storage/Handler/RedisSessionHandler.php
+++ b/Session/Storage/Handler/RedisSessionHandler.php
@@ -270,7 +270,16 @@ LUA;
     /**
      * Shutdown handler, replacement for class destructor as it might not be called.
      */
-    public function shutdown() {
+    public function shutdown()
+    {
         $this->close();
+    }
+
+    /**
+     * Destructor.
+     */
+    public function __destruct()
+    {
+        $this->shutdown();
     }
 }


supersmile2009:
Well, I re-checked it, and it seems that I looked at the wrong test run when opening this PR.
My PR doesn't solve the problem.
Reverting PR #348 solves issue completely.

I guess the problem is with registering shutdown function. When you boot many kernels in a big test suite it register this function every time, wasting up memory.

I think we need a config option to optionally enable/disable registering shutdown function. I'll update this PR tomorrow with such implementation.

curry684:
Seems related to #400?

But isn't the root cause of this (and that) problem that you're actually trying to execute all those tests in the same process, which is something that PHP is really not meant to do given the whole existence of functions like `register_shutdown_handler`?

I'm curious whether your issues are solved if you run phpunit with `--process-isolation=true`, or add `processIsolation="true"` to the XML config file, or use annotations like `@runInSeparateProcess`. See also [`@runTestsInSeparateProcess`](http://phpunit.readthedocs.io/en/7.1/annotations.html?highlight=runinseparateprocess#runtestsinseparateprocesses) on how to apply this to a single test class in the suite.

Keep in mind that we will likely not consider it wise to change/break perfectly fine code for web request handling scenarios just to facilitate unit tests, if the tests can be easily fixed by configuration or annotations that were invented for this precise purpose.

Seldaek:
We could maybe merge this and then enable the register_shutdown_function behavior only when locking is enabled. Without locking, a request timing out/failing probably shouldn't write the session data anyway. The problem when locking is that you end up with a session deadlock and block other requests for a while.

curry684:
Possibly more practical approach: we could use [`framework.test` setting](https://symfony.com/doc/current/reference/configuration/framework.html#test) to disable/enable some behaviors, like not setting shutdown handlers on one hand and aggressively closing open connections during kernel termination otherwise. Would be a matter of just compiling some extra event handlers in there.

Seldaek:
Yes.. or just making the register_shutdown_function optional with a constructor argument, then it's easy to toggle on/off. Because down the line this is a bad idea, it prevents writing app servers for example without leaking memory.

curry684:
Quick fix for 2.1 could also be to introduce some static behaviors, pseudocode:
```
static $cleanup = [];
if (empty($cleanup)) {
  register_shutdown_handler(function() use ($cleanup) {
    foreach ($cleanup as $that) {
      $that->close();
    }
  });
}
$cleanup[] = $this;

**Comment by Seldaek** [comment]
I don't know how much this helps, it'll still keep a handle to every instance created so that leaks all the same IMO.

**Comment by curry684** [comment]
Mmm true. So we should likely merge this and keep the `framework.test` approach in mind for a more structural solution later.

**Comment by supersmile2009** [comment]
@curry684 Thanks for the tip about `--process-isolation`, I wasn't familiar with this option. It may work, but at the moment enabling it breaks a lot of out tests, so we'd rather use a patch in our composer.json for now to disable shutdown function.

As suggested by @Seldaek, register_shutdown_function should only be enabled when session locking is enabled.
For the moment apart from what you see in this PR, we're going to use a small patch adding this
```
        if (true === $locking) {
            register_shutdown_function(array($this, 'shutdown'));
        }
```
into RedisSessionHandler constructor.
This way it's possible to set `snc_redis.session.locking: false` in test config, which will solve the issue.
I'm not adding it into this PR since you already approved it, but If you think it's a good idea, I can push it to this branch.

**Comment by curry684** [comment]
> enabling it breaks a lot of out tests

That indicates your tests depend on order of execution and/or other tests to succeed, which is considered [bad practice](http://essentials.xebia.com/independent-tests/). Interdependency should be handled with `@depends` annotations, making a chain of tests explicit.

> I can push it to this branch

Fine with me, looks harmless enough and I'll merge it as it's a quick workaround for now.

**Comment by curry684** [comment]
For the record: holding off on merge until that addition is made.

**Comment by curry684** [comment]
Also #400 is a slightly different problem that we're not solving now. Could we do that in a practical way?

**Comment by curry684** [comment]
Merged, thanks!

**Comment by goetas** [comment]
Wanted just to share here my experience with session locking and Redis. 
File based sessions use the filesystem locking, so if PHP process that is holding the lock terminates abnormally, it will automatically release the lock. This is also true if the script reaches the max_execution_time and phpfpm decides to terminate the child (it is done sending a SIGTERM signal).
Unfortunately when using redis based locking, is fundamental that the script terminates gracefully, so the destructor gets called and release the lock. Unfortunately that does not happen, and some session are locked.

As a solution we have used:

```php
pcntl_async_signals(true);
pcntl_signal(SIGTERM, static function () {
    exit;
});
```

That installs a signal handler for the `SIGTERM` signal and uses `exit`. I this way when PHP-FPM sends the termination signal, a normal `exit` is executed, that will allow PHP to invoke all the session unlocking features.

This implementation has as side effect that the worker process does not terminate on `SIGTERM` anymore because `exit` in reality does not terminate the FPM worker, but just tells to PHP to stop the script execution.

Does anyone have an idea on how to handle this in a better way?

B-Galati:
@goetas The Symfony Lock component may have fixed this problem a better way?

goetas:
@B-Galati I do not think that is possible as locking issues are caused by phpfpm killing the process without calling destructors

Issue #356, opened 2017-08-28
RedisEnvDsn will break with Symfony 3.4/4.1


nicolas-grekas:
In https://github.com/snc/SncRedisBundle/blob/master/DependencyInjection/Configuration/RedisEnvDsn.php#L25

you hardcode the env placeholder regexp. This is an internal detail that is not covered by the BC policy and will change in 3.4 - thus the logic will break then.

The correct logic should use `$container->resolveEnvPlaceholders()` instead.

xawiers:
+1

dzubchik:
I've spent 2 hours debugging this! I was looking for errors in logstash configuration, monolog and event my DNA.

birkof:
+1

nicolas-grekas:
The break will be removed by https://github.com/symfony/symfony/pull/24486
(but using resolveEnvPlaceholder would still be a good idea.)

nicolas-grekas:
Reopening: version 4.1 did change the pattern of env placeholders. This bundle will break unless this is fixed.

curry684:
Hrm thanks for spotting it, we really should reconsider the entire mechanism for 3.x of this bundle.

B-Galati:
I will try to fix this in coming weeks for 2.1.

curry684:
Still an issue in 2.1.3

PR #413, opened 2018-04-13
Fix support for Heroku style REDIS_URL env variables

### LINKED ISSUE #410: [Bug] DSN is not interpolated at runtime with environment variable
Hello,

I just upgraded from 2.0.6 to 2.1.0 and my app broke because the DSN was not updated.

It looks like the issue is similar to https://github.com/snc/SncRedisBundle/issues/302.

I checked my cache and the value of the config is hard-coded instead of using the env variable.

Here is my config :

```
snc_redis:
    clients:
        default:
            type: predis
            alias: default
            dsn: "%env(REDIS_DSN)%"
    session:
        client: default
        prefix: session
```

Am I missing something ?

In the meantime I moved back to 2.0.6 and everything is back to normal.

Let me know for anything and thanks for your help.

B-Galati:
Basically re-add what was done in #353

fixes #410 

I also think that redis profile should not be resolved/interpolated environment variable in cache.


curry684:
Can you rebase on current master? It should fix the tests.

B-Galati:
@curry684 Done and also make some changes, thanks ;-) 

It looks good overall but I am not sure if it's breaking something else or not.

curry684:
I think tests should catch that, and it's going into master not right into stable.

B-Galati:
Thanks. What should I do to fix 2.1?

thomasbisignani:
Good job @B-Galati ! :medal_sports:

curry684:
Merged it into master, not bothering with 2.1 anymore as it's pretty invasive and I'm trying to keep that one stable after recent issues.

curry684:
Hmm my bad, I thought it was broken before 2.1.0 already. I'll see if I can merge it back into 2.1.

curry684:
See #356, we're reopening up an old issue here.

PR #415, opened 2018-04-18
Close connections on shutdown

supersmile2009:
This PR addresses issue #400.
`SncRedisBundle::shutdown()` function will be called on kernel shutdown in-between unit tests. The same solution is implemented in DoctrineBundle.

I also had to declare `snc_redis.session.handler` as a public service to be able to fetch it from container in Symfony 4+.

### DIFF
--- a/Resources/config/session.xml
+++ b/Resources/config/session.xml
@@ -11,7 +11,7 @@
     </parameters>
 
     <services>
-        <service id="snc_redis.session.handler" class="%snc_redis.session.handler.class%">
+        <service id="snc_redis.session.handler" class="%snc_redis.session.handler.class%" public="true">
             <argument type="service" id="snc_redis.session.client" />
             <argument>%session.storage.options%</argument>
             <argument>%snc_redis.session.prefix%</argument>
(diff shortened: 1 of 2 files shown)


curry684:
Solution is fine, but I don't think we should be making a service public just for testing purposes. Don't we have some way to do that better?

supersmile2009:
This is the most elegant and simple solution as to my knowledge.

There is a way to fetch private services from container in tests env, but it's Symfony 4.1 feature:  [link1](https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing), [link2](https://github.com/symfony/symfony/pull/26499)

There is a service resetting feature in Symfony 3.4+ https://github.com/symfony/symfony/pull/24155, but it runs on kernel.terminate event, which is not the case with kernel shutdown.
We could reproduce similar feature here by extending a listener from above mentioned Symfony's PR, adding a compiler pass for it, adding some kind of 'kernel.shutdown' event. But it will require people using this bundle to override `Symfony\Bundle\FrameworkBundle\Test\KernelTestCase::ensureKernelShutdown()` in their tests to dispatch that event in order to benefit from that feature. Too much effort with not a really good result.

The perfect solution would probably be adding existing service resetting feature into kernel shutdown - but that's an issue to be discussed in Symfony repo. And I guess it's going to be Symfony 4.1 feature, I don't think they'll merge it into 4.0.

dkarlovi:
You can add an env-specific service and prefix it with `test.`.

curry684:
I'll merge this into 2.1 but keep #400 open so we can fix it better in 3.x.

Issue #267, opened 2016-05-06
DSN defination problem: Cluster hostname


faruktemur:
Hi @snc ,

We successfuly connect to redis when we define hostname of all 3 nodes in redis cluster.

We are also connecting to redis cluster when we define directly AWS LoadBalancer's hostname to DSN. But, client is not catching MOVED exceptions. As we understand client is not connecting as cluster mode if we define only 1 hostname.

my config.yml

```
snc_redis:
    clients:
        cluster:
            type: predis
            alias: cluster
            logging: "%kernel.debug%"
            dsn: ['redis://example.eu-west-1.elb.amazonaws.com']
            options:
                profile: 2.4
                connection_timeout: 10
                connection_persistent: true
                read_write_timeout: 30
                iterable_multibulk: false
                throw_errors: true
                cluster: "redis-cluster"
                replication: false
```

fthyilmz:
#269

snc:
The cluster option expects a class name or callable (see https://github.com/nrk/predis/wiki/Client-Options). Your config contains `cluster: "redis-cluster"` which looks weird.

fthyilmz:
Hi @snc 

You are passing string to predis client directly without any validation. Thus, "**redis-cluster**" option creates instance of RedisCluster class as you can see in below.

The cluster options: [See](https://github.com/nrk/predis/blob/v1.1/src/Configuration/ClusterOption.php#L35-L49)

This was our hack. 

Actually, our main problem is not creating RedisCluster instance with only one hostname because of statement in predis [See](https://github.com/nrk/predis/blob/v1.1/src/Client.php#L109)

In our application there is a Load Balancer in front of all Redis Cluster. That's why we have added cluster option statement in pull request.

ebarault:
> Actually, our main problem is not creating RedisCluster instance with only one hostname because of statement in predis

@fthyilmz: finally did you find a solution to this problem? i saw you closed your PR #269 

I'm trying to use aws's Redis which comes with a single "configuration endpoint" which is somewhat similar to your load balander in front of your cluster, but i'm having "MOVED" except

curry684:
Fixed by #416

PR #416, opened 2018-04-18
Create cluster connection correctly for one host

### LINKED ISSUE #402: AWS Elasticache Redis Cluster : Uncaught PHP Exception Predis\Response\ServerException: MOVED
Hi, 

AWS provides a single configuration endpoint for an Elasticache Redis Cluster (x shards, y replicas per shards).

With the following configuration, i get this error: 
```
Uncaught PHP Exception Predis\Response\ServerException: "MOVED 2777 w.x.y.z:6379" at /var/www/html/vendor/predis/predis/src/Client.php line 370
```

```
snc_redis:
    clients:
       default:
         type: predis
         alias: default
         dsn:
           - my-redis.cmfj2h.ng.0001.euw3.cache.amazonaws.com:6379
         options:
           profile: 3.2
           cluster: redis
           replication: false
```

I use snc_redis bundle **2.0.4** and predis bundle **1.1.1**

I find very little doc (mostly in issues) about server-side replication/sharding support option (`cluster: redis`).

Has anyone solved this?

linasm83:
Fixes #402, #267 (might also fix some other clustering issues when only one dsn is defined)

If only one dsn is defined when cluster option is passed connection instance class is incorrect (should be [Predis\Connection\Aggregate\RedisCluster](https://github.com/nrk/predis/blob/v1.1/src/Connection/Aggregate/RedisCluster.php).

The problem is that we're passing Predis\Connection\ParametersInterface as first argument to client, but we should pass array instead cause only then [Predis\Client::createConnection](https://github.com/nrk/predis/blob/v1.1/src/Client.php#L109) method will create correct connection class.

### DIFF
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -135,6 +135,8 @@ class SncRedisExtension extends Extension
     {
         if (null === $client['options']['cluster']) {
             unset($client['options']['cluster']);
+        } else {
+            unset($client['options']['replication']);
         }
 
         // predis connection parameters have been renamed in v0.8
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -203,7 +203,7 @@ class SncRedisExtension extends Extension
         $clientDef = new Definition($container->getParameter('snc_redis.client.class'));
         $clientDef->setPublic(true);
         $clientDef->addTag('snc_redis.client', array('alias' => $client['alias']));
-        if (1 === $connectionCount) {
+        if (1 === $connectionCount && !isset($client['options']['cluster'])) {
             $clientDef->addArgument(new Reference(sprintf('snc_redis.connection.%s_parameters.%s', $connectionAliases[0], $client['alias'])));
         } else {
             $connections = array();


curry684:
Seems simple enough, since I don't have clusters to test on hoping we can get confirmation of the fix from people that experienced issues here!

curry684:
@linasm83 do you understand the issues reported at https://github.com/snc/SncRedisBundle/issues/402#issuecomment-383520254?

linasm83:
@curry684 I think it should only work`replication: false` if setting `replication: true` then you need to provide at least two dsn's for predis client (one master / multiple slaves).

Looking at predis client class it looks it only supports cluster or replication, but not both at the same time.

So not sure there's proper solution on that case, maybe only some additional config validation not allow to pass only one dsn if `replication: true`?

curry684:
Makes sense that you cannot use both at the same time, would cause awful race conditions I suppose. In that case yes we should fail early during compilation.

@ebarault what's your take on this?

ebarault:
The purpose of backend-side-managed clustering is to get rid of any extra configuration at client side.
Hence we could completely drop the requirement for the `replication` param when cluster type `redis` is selected.

This config should fail with an error early and with the appropriate error message.

ebarault:
this is with predis version 1.1.1 (latest released)

linasm83:
@ignaciorivmen try to remove `replication: false` at all

```
snc_redis:
    clients:
        default:
           type: predis
           alias: default
           dsn:
               - %redis_dsn_1%
           options:
               profile: 3.2
               cluster: redis
    session:
        client: default
        prefix: "%session_prefix%"
```

maybe we should really ignore `replication` parameter when `cluster: redis` is defined

ebarault:
bingo: we tried without the replication param and now the app compiles
the app seems to behave normally, but we get these errors in symfony logs:

```
snc_redis.ERROR: Command "SET session_prefix_dev135c5b319c119c6e6ce8da0524a8ab9d.lock 5ae0d948658e9 PX 30001 NX" failed (MOVED 7537 172.50.197.202:6379)
```

in other words, the storage of sessions in redis seems to fail

linasm83:
seems strange as also storing sessions in aws redis cluster, and my config is similar to provided, just not passing profile option...

curry684:
Meh the wrappers are causing a lot of issues considering they were written ages before PHP started supporting type safety in function arguments.

ebarault:
I'm not sure which additional test could be ran to bring more food for thoughts.
From a black-box perspective, i'd say the fix works.

curry684:
> maybe we should really ignore replication parameter when cluster: redis is defined

Sounds like a good idea if that causes the fix to be universally applicable, could you add that?

curry684:
Thanks!

ebarault:
@curry684 : hi, could you tag a new release and propagate this fix to composer ?

curry684:
Yes, been running 2.1.x for a few days myself now in current project and it seems all stable.

curry684:
https://github.com/snc/SncRedisBundle/releases/tag/2.1.3

ebarault:
thanks 👍

PR #423, opened 2018-05-02
Command executions should be logged as DEBUG instead of INFO.

stephanvierkant:
…ems where INFO is being used to log informative messages like 'user logged in'.A loop with redis for e.g. in a background-process will pollute the logs extremely.

(cherry picked from commit af2e3f9)

I wanted to make this change, but I found this commit in the master branch. That brings me to another question: what's the reason behind 4 different branches (master, 1.0, 1.1, 2.1)?

### DIFF
--- a/Logger/RedisLogger.php
+++ b/Logger/RedisLogger.php
@@ -62,7 +62,7 @@ class RedisLogger
                     $this->logger->err($message);
                 }
             } else {
-                $this->logger->info('Executing command "' . $command . '"');
+                $this->logger->debug('Executing command "' . $command . '"');
             }
         }
     }
--- a/Tests/Logger/RedisLoggerTest.php
+++ b/Tests/Logger/RedisLoggerTest.php
@@ -45,7 +45,7 @@ class RedisLoggerTest extends TestCase
 
         $this->logger
             ->expects($this->once())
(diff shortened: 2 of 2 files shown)


curry684:
Heya! We consciously decided not to merge this change into 2.1 as [technically it's a sort of breaking change](https://github.com/snc/SncRedisBundle/pull/329#issuecomment-377321211). I agree the impact is not that big, but neither is having some extra logging crop up during development (all logging is disabled by default in production anyway).

The branches are a pretty common development workflow adhering to [SemVer practices](https://semver.org/). `master` is where active development is done, currently aliased to `3.0` as well as that'll be the next major release. `2.1` allows us to separately release patches in the `2.1.x` range. The `1.0` and `1.1` branches are technically obsolete, but do no harm.

If you look at [Packagist](https://packagist.org/packages/snc/redis-bundle) you can also see how the repository tags correspond to actual releases (`2.1.1`, `2.0.5` etc.) and branches are also accessible (`2.1.x-dev` corresponds to our 2.1 branch).

curry684:
Thanks for the effort, but I'm closing this as we can't merge into 2.x for given reasons and it's already done in 3.x.

Issue #253, opened 2016-03-10
Disable bundle if redis server down / not installed


trousers:
At the moment if the redis server goes down the application will crash / display a 503 error

Also for environments such as test environments there is no way (or I can't find a way) to disable the bundle

Is it not possible to

1. option to disable the bundle in the environment config

2. detect if no server running disable the bundle

trousers:
The problem is that if redis is down the whole application is down, so the solution would be to clear the cache once it's available again. Auto disable / manual enable?

pounard:
The auto disable problem is the same if you have stalled data in your legacy backends, I think that there is no magical solution to this problem, as the Drupal Redis module maintainer, this very same problem has been a long standing discussion, many months actually, and a few people did some custom fallback by hacking the code, but no viable generic solution was ever found to answer to this problem.

trousers:
I'm using doctrine, so the solution was to use the doctrine cache bundle instead. 
If redis down, use filecache with a "redis_" key. Then if the redis filecache key exists and redis is up delete both redis and filecache.

mmucklo:
It's possible to disable this bundle in test via config.yml trickery (just put the snc_redis section in config_prod.yml, but not in config.yml, for example, using the symfony standard layout).

Nevertheless does this bundle support sentinel?

trousers:
@mmucklo 

How does that work? the bundle is registered in athe appkernel and asks for the configuration, specifically:

`The child node "clients" at path "snc_redis" must be configured.`

mmucklo:
Ah correct, my mistake I started to add about AppKernel, but thought the config trick would be enough.

See line 21 here for a hint on how to turn on/off bundles depending on environment (in your case you'd want to do the inverse - only add the bundle if the environment is not "test"):

https://github.com/symfony/symfony-standard/blob/f5b1ae88c3d2338f43a18363811f03bae196ff54/app/AppKernel.php#L21

The only problem is if any of your then-testedr code goes to use Redis it may fail looking for a non-existent service if everything is hard-wired - though you can "hack" around that as well.

clytemnestra:
I'm also looking for something similar. It makes no sense for the whole app to crash when all the data would be available at the original source (hard disk). There should be an option to allow the app to function as if Redis wouldn't exist, if the connection to Redis fails. 

How can it be done?

luispabon:
I see two separate issues:
[...]

PR #424, opened 2018-05-07
Add on demand phpredis connections

borisbabic:
This extends the wrapper to allow connections to be seamlessly created only when necessary.

### DIFF
--- a/Client/Phpredis/Client.php
+++ b/Client/Phpredis/Client.php
@@ -36,22 +36,65 @@ class Client extends Redis
     protected $alias;
 
     /**
+     * Whether an on demand connection is still being awaited
+     *
+     * @var bool
+     */
+    protected $onDemandPending = false;
+
+    /**
+     * An array of delayed calls arrays for on demand connections
+     * A delayed call array has two keys, "method" - method name and "args" - an array of arguments;
+     * @var array
+     */
+    protected $delayedCalls = array();
+
+    /**
      * Constructor.
      *
-     * @param array       $parameters List of parameters (only `alias` key is handled)
+     * @param array       $parameters List of parameters (only `alias` and `on_demand_connect` keys are handled)
      * @param RedisLogger $logger     A RedisLogger instance
      */
     public function __construct(array $parameters = array(), RedisLogger $logger = null)
     {
         $this->logger = $logger;
         $this->alias = isset($parameters['alias']) ? $parameters['alias'] : '';
+        $this->onDemandPending = isset($parameters['on_demand_connect']) ? $parameters['on_demand_connect'] : false;
     }
+
+    /**
+     * Delays the method call until an ondemand connection is established
+     *
+     * @param string $methodName
+     * @param array $arguments
+     */
+    private function delay($methodName, array $arguments)
+    {
+        if ($this->onDemandPending) {
+            $this->delayedCalls[] = array("method" => $methodName, "args" => $arguments);
+        } else {
+            call_user_func_array("parent::$methodName", $arguments);
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


curry684:
I'm sorry but this PR is currently by definition broken, as it depends on the connection wrapper class which is fundamentally broken right now (#399).

Also I'm not sure I see the use case for the ton of complexity this PR is adding. The connection itself is already lazily instantiated at the Symfony service level, so afaics there would only be a difference if you actually inject the connection somewhere and not actually use it. In all other cases it would at best be a micro-optimization, at worst a ton of tech debt waiting to become a problem for no discernible performance win.

Could you perhaps elaborate on the use case and the real world problem you intend to solve?

borisbabic:
> I'm sorry but this PR is currently by definition broken, as it depends on the connection wrapper class which is fundamentally broken right now (#399).

I wasn't aware of that issue, that's very problematic

> Also I'm not sure I see the use case for the ton of complexity this PR is adding. The connection itself is already lazily instantiated at the Symfony service level, so afaics there would only be a difference if you actually inject the connection somewhere and not actually use it. In all other cases it would at best be a micro-optimization, at worst a ton of tech debt waiting to become a problem for no discernible performance win.
>
>Could you perhaps elaborate on the use case and the real world problem you intend to solve?

This wasn't really for performance but rather error handling. It came up at work where we want to make a critical monolithic service more robust. We have many situations where we can continue on if there are any redis related issues. If the client is injected as is we don't get any chance to handle connection errors and continue depending on what was attempted.

The following is a simplistic example of what we'd like to do:

```php
<?php

class Cache
{
    private $redis;

    public function __construct(\Redis $redis)
    {
        $this->redis = $redis;
    }

    public function get($key)
    {
        try {
            return $this->redis->get($key);
        } catch (\Throwable $_) {
            // if there are any redis issues, including connection issues, we want to continue
            return null;
        }
    }
    
    public function set($key, $value)
    {
        // we don't want to allow the old value to silently remain so we want an error here
        $this->redis->set($key, $value);
    }
}
```

curry684:
So your issues are essentially similar to #105 and #253 ?

borisbabic:
Yes, it's similar. Basically we also want to somehow handle redis connection issues. 

To me there seems to be a difference in the approach; we want to have the ability to decide how to handle issues ourselves rather than having the bundle do it.

curry684:
For the time being I'm highly hesitant to merge this PR as it's really invasive, a highly specific implementation for your case, and currently downright broken  😉 

However what you may not be aware of is that you can implement this yourself - the bundle config allows you to override the wrapper class yourself (https://github.com/snc/SncRedisBundle/blob/master/DependencyInjection/Configuration/Configuration.php#L53). This should allow you to implement your solution locally without effort while we implement a more generic solution in v3 of the bundle, which will also be fully phpredis 4 compatible out of the box then.

borisbabic:
This PR actually started as an override to the wrapper class. At first I wanted to just extend the wrapper but I  had to resort to basicaly copy/pasting it and in the end it seemed better to make this configurable and integrate it upstream.

I don't think the invasiveness is that big of an issue. The really invasive part is opt in and, IMHO, can be sufficiently documented.

I also don't think it's actually too specific. Those involved in the other issues may have wanted to solve their problems differently but it seems to me that they could have used this to solve their issue.

However I do understand not merging this because of #399, the wrapper seems broken.

If this won't be merged would you be open to making it possible to extend the wrapper with this functionality? Duplicating all the redis methods in a new class in our project seems like a headache waiting to happen.

curry684:
Closing in favor of #440

PR #440, opened 2018-06-30
#434 Enable Lazy service for phpredis

### LINKED ISSUE #434: Support pre-building the Symfony container without the Redis server available (or known)
In containerized apps, it's common to pre-warm cache while building the image. This allows for the container to come in with warm cache and just start working as if it's here forever.

With SncRedisBundle, this is currently not possible as it seems the bundle is trying to connect to the Redis server at build time. But, not only the Redis server might not be available at build time, its connection string is also not known (supplied later via env).

```yaml
parameters:
    env(REDIS_URL): ""

snc_redis:
    clients:
        default:
            type: phpredis
            alias: default
            dsn: "%env(REDIS_URL)%"
        cache:
            type: phpredis
            alias: cache
            dsn: "%env(REDIS_URL)%"
        session:
            type: phpredis
            alias: session
            dsn: "%env(REDIS_URL)%"
    session:
        client: session

framework:
    session:
        handler_id: snc_redis.session.handler
    cache:
        app: cache.adapter.redis
        default_redis_provider: snc_redis.cache
```

This config doesn't allow the container to be built. First the extension claims the URL is invalid:
https://github.com/snc/SncRedisBundle/blob/38de7edaeacaa1c8ce5aa5fa1eba7697c6c051d9/DependencyInjection/SncRedisExtension.php#L110

B-Galati:
The service will be lazy if `symfony/proxy-manager-bridge` is installed
As a consequence it will allow to warmup the cache even if Redis server is not reachable

fix #434

### DIFF
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -276,6 +276,7 @@ class SncRedisExtension extends Extension
         $phpredisDef->addArgument($client['alias']);
         $phpredisDef->addTag('snc_redis.client', array('alias' => $client['alias']));
         $phpredisDef->setPublic(false);
+        $phpredisDef->setLazy(true);
 
         $container->setDefinition($phpredisId, $phpredisDef);
         $container->setAlias(sprintf('snc_redis.%s', $client['alias']), new Alias($phpredisId, true));
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -386,3 +386,14 @@ snc_redis:
         client: profiler_storage
         ttl: 3600
 ```
+
+## Troubleshooting ##
+
+If cache warmup fails for prod because a redis server is not available,
+try to install [`symfony/proxy-manager-bridge`](https://symfony.com/doc/master/service_container/lazy_services.html):
+
+``` bash
+$ composer require symfony/proxy-manager-bridge
+```
+
+Once done some services will be lazy-loaded and could prevent unwanted connection call. 
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
     "suggest": {
         "monolog/monolog": "If you want to use the monolog redis handler.",
         "predis/predis": "If you want to use predis.",
-        "symfony/console": "If you want to use commands to interact with the redis database"
+        "symfony/console": "If you want to use commands to interact with the redis database",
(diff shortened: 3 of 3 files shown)


B-Galati:
@dkarlovi Would you be able to test is out ? (or you can just edit your vendors with this little change)

B-Galati:
I guess I should also edit the doc about `symfony/proxy-manager-bridge` and phpredis behavior

B-Galati:
Doc updated

B-Galati:
@dkarlovi What do you wean exactly ? 

From my knowledge, I do think lazy service is a good solution as it's something easy, powerful and well supported by Symfony.

dkarlovi:
@B-Galati I mean, would making the service lazy have the desired behaviour and/or bad consequences elsewhere?

From discussions at Symfony's Slack, `lazy=true` is almost considered an anti-pattern and (some) users called for the feature to be completely deprecated by Symfony.

B-Galati:
@dkarlovi Thank you. Can you give the reasons behind this thinking. What would be the alternatives?

I am not aware about any bad consequences for lazy services. Certainly we gonna lose some very little performances in some situation but that's all I can see. In other situation where for example, the redis client is not used like the cache warmup, performances will be better.

IMHO lazy services are just an infrastructure detail.

Perhaps I am wrong so please let me know for anything.

OskarStark:
Symfony just decided to make all doctrine listeners lazy by default in 4.2

dkarlovi:
@B-Galati sorry for the delay. This does indeed fix #434 (as in, allows the container to be prebuilt). There doesn't seem to be any side-effects that are noticeable to me.

About lazy services, this might have been a "storm in a teacup" by a group of loud users, but the prominent opinion was lazy services are almost always a bad idea. This might be considered an exception. :)

Thank you for working on this!

B-Galati:
@dkarlovi Thanks for the feedback :-)

Basically PR #424 is doing the same thing but the implementation is different.

dkarlovi:
@curry684 I see you've merged this, would you consider a bugfix release?

piotrkochan:
Lazy loading for this service makes PHP 5.6 fail because of invalid cached Redis class. See https://github.com/snc/SncRedisBundle/issues/465

Issue #453, opened 2018-08-22
PhpRedis Cluster


furester:
Since there is no client implementing RedisCluster or RedisArray from PhpRedis is it possible (and how) to configure a cluster with this library instead of predis?

curry684:
There is currently no support whatsoever on our end for `RedisCluster` and `RedisArray`.

furester:
ok, thanks, i can try to manage `RedisCluster` support, 'cause I'm actually using it with old silex projects. I'll send a `pull request` then.

furester:
i'll implement a common interface in `Client\Phpredis` to let developers switch between `\Redis` and `\RedisCluster` using only configuration files, but here it came first problem: these classes has methods with same name (`info` for example), but different footprint (`\Redis` and `\RedisCluster` doesn't implement a common interface - ref. https://github.com/phpredis/phpredis/issues/1298).

In the first implementation i'll comment out these methods to avoid conflicts, is it suitable for you?

PR #454, opened 2018-08-23
Add Support for Phpredis\RedisCluster (#453)

furester:
This shoudl fix issue #453 with a couple of constraints:

Added a configuration parameter to switch between RedisCluster implementations (`snc_redis.phpredis_clusterclient.class`) and based on `cluster` option in client configuration switch between `\Redis` and `\RedisCluster`.

`Snc\RedisBundle\Client\Phpredis\ClientInterface` is implemented by both `Snc\RedisBundle\Client\Phpredis\Client` and `Snc\RedisBundle\Client\Phpredis\ClientCluster`, but directed node commands are commented out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ matrix:
     - php: nightly
     - env: SYMFONY_VERSION="4.4.*" SYMFONY_DEPRECATIONS_HELPER="weak" STABILITY="dev"
 
+dist: xenial
 sudo: false
 
 cache:
@@ -39,15 +40,17 @@ env:
 
 before_install:
   - phpenv config-rm xdebug.ini || true
-  - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi;
+  - if [[ ${TRAVIS_PHP_VERSION} != hhvm ]]; then echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi;
   - php -i | grep -i --color=never "Redis Version"
-  - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then phpenv rehash; fi;
+  - if [[ ${TRAVIS_PHP_VERSION} != hhvm ]]; then phpenv rehash; fi;
   - composer self-update
-  - if ! [ -z "$STABILITY" ]; then composer config minimum-stability ${STABILITY}; fi;
-  - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update symfony/symfony:${SYMFONY_VERSION}; fi;
+  - if ! [ -z "${STABILITY}" ]; then composer config minimum-stability ${STABILITY}; fi;
+  - if [ "${SYMFONY_VERSION}" != "" ]; then composer require --no-update symfony/symfony:${SYMFONY_VERSION}; fi;
+  - for PORT in {7000..7008}; do sudo redis-server --port ${PORT} --cluster-enabled yes --cluster-config-file ${PORT}.conf --daemonize yes; echo 127.0.0.1:${PORT}; done
+  - echo yes | redis-cli --cluster create $(seq -f 127.0.0.1:%g 7000 7008) --cluster-replicas 2
 
 install:
-  - composer update --no-interaction --no-progress $COMPOSER_FLAGS
+  - composer update --no-interaction --no-progress ${COMPOSER_FLAGS}
 
 before_script:
   - mkdir Snc && ln -s ../ Snc/RedisBundle
--- a/Client/Phpredis/ClientCluster.php
+++ b/Client/Phpredis/ClientCluster.php
@@ -1,29 +1,16 @@
 <?php
 
-/*
- * This file is part of the SncRedisBundle package.
- *
- * (c) Henrik Westphal <henrik.westphal@gmail.com>
- * (c) Yassine Khial <yassine.khial@blablacar.com>
- * (c) Pierre Boudelle <pierre.boudelle@gmail.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
 namespace Snc\RedisBundle\Client\Phpredis;
 
-use Redis;
+use RedisCluster;
 use Snc\RedisBundle\Logger\RedisLogger;
 
 /**
- * PHP Redis client with logger.
+ * PHP RedisCluster client with logger.
  *
- * @author Henrik Westphal <henrik.westphal@gmail.com>
- * @author Yassine Khial <yassine.khial@blablacar.com>
- * @author Pierre Boudelle <pierre.boudelle@gmail.com>
+ * @author Igor Scabini <furester@gmail.com>
  */
-class Client extends Redis
+class ClientCluster extends RedisCluster
 {
     /**
      * @var RedisLogger
@@ -41,10 +28,12 @@ class Client extends Redis
      * @param array       $parameters List of parameters (only `alias` key is handled)
      * @param RedisLogger $logger     A RedisLogger instance
      */
-    public function __construct(array $parameters = array(), RedisLogger $logger = null)
+    public function __construct(array $seeds = array(), array $parameters = array(), RedisLogger $logger = null)
     {
         $this->logger = $logger;
-        $this->alias = isset($parameters['alias']) ? $parameters['alias'] : '';
+        $this->alias = $parameters['alias'] ?? '';
+
+        parent::__construct(null, $seeds);
     }
 
     /**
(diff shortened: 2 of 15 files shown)


curry684:
Thanks! Before we go move on to reviewing - did you notice your new tests are broken on PHP < 7.2?

furester:
Thank you for pointing it to me, using phpredis 4.1.1 some methods are defined with php 7.2 syntax, for example:

> public function zUnion($key, array $keys, ?array $weights = NULL, $aggregate = NULL)

and this is causing the problem, i'll try to find a solution .. or is there any suggested solution?

curry684:
Hey @furester, as we're moving to stabilize 3.0 (currently in RC2) I think it's a good idea to update this code so it can still make release. Any chance of you being able to pick this up?

.travis.yml
```diff
@@ -23,6 +23,8 @@ matrix:
       env: SYMFONY_VERSION="4.3.*" SYMFONY_DEPRECATIONS_HELPER="weak" STABILITY="dev"
     - php: nightly
   allow_failures:
+    - php: 7.0
```
B-Galati: Please revert these 2 lines

Client/Phpredis/ClientInterface.php
```diff
@@ -0,0 +1,927 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Snc\RedisBundle\Client\Phpredis;
+
+use Redis;
+use Snc\RedisBundle\Logger\RedisLogger;
+
+/**
```
B-Galati: You can remove these docblock entirely

Client/Phpredis/ClientInterface.php
```diff
@@ -0,0 +1,927 @@
+<?php
+
+/*
```
B-Galati: You can remove these docblock entirely

Client/Phpredis/ClientCluster.php
```diff
@@ -0,0 +1,1545 @@
+<?php
+
+/*
```
B-Galati: You can remove these docblock entirely

Client/Phpredis/ClientCluster.php
```diff
@@ -0,0 +1,1545 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Snc\RedisBundle\Client\Phpredis;
+
+use RedisCluster;
+use Snc\RedisBundle\Logger\RedisLogger;
+
+/**
```
B-Galati: You can remove these docblock entirely

Client/Phpredis/ClientCluster.php
furester: I think i didn't understand this suggestion, meantime pushed a new revision

Client/Phpredis/ClientCluster.php
furester: I think i didn't understand this suggestion, meantime pushed a new revision

furester:
@B-Galati I pushed an update revision with most of your suggestions and removing ClientInterface, - main purpose of a common interface is to type-hint arguments, so we could not manually wire every class.

We could also use an "empty" interface, what do you think about this?

B-Galati:
@furester the empty interface would not be that useful IMHO as it would not allow to know which methods are available. 

Also with Symfony binding you can autowire any service based on a variable name so that would do the job if one wants to go this way.

I'll do another review ASAP :-)

B-Galati:
@furester I just tried it locally and push few changes on your branch. I encountered the following issue -> https://github.com/phpredis/phpredis/issues/1532 and don't know how to fix it.Any idea? 

Also does RedisCluster implies the use of RedisArray? Cause it's not possible with the current implementation.

Could you update the doc and tests?

Do you have an example of configuration that you are using?

FYI my phpredis version is 3.1.6.

furester:
@B-Galati actually this is my `snc_redis.yml` config file:

```snc_redis:
    class:
        phpredis_client: Snc\RedisBundle\Client\Phpredis\Client
        phpredis_clusterclient: Snc\RedisBundle\Client\Phpredis\ClientCluster
        
    clients:
[..]
        cache:
          type: phpredis
          alias: cache
          dsn: ["redis://%env(REDIS_CACHE_URL)%/0"]
          logging: false
          options:
              profile: 4.1
              prefix: "lost-basket:"
              cluster: "phpredis-cluster"
              replication: false
```

where `REDIS_CACHE_URL` is in the form `redis://clustername.c1008j.clustercfg.az.cache.amazonaws.com`
as you notice i am using aws elasticache with mode cluster enabled.

speaking about seeds issue i am working with phpredis version from 4.1 to 5.0, but i don't think that it's a lib version problem, could you share your configuration and environment details?

In phpredis RedisArray is an other class without common interface as before so it needs a different implementation and one is not dependent on the other, atm i haven't a redis array to test with. I suggest to end with this one and then open a new issue to add RedisArray support if needed.

I'll go forward with doc and tests.

furester:
@B-Galati if your error is something like this one:
https://travis-ci.org/snc/SncRedisBundle/jobs/566892255
it could be possible that redis isn't configured with cluster mode enabled (as in travis test that i'm going to revert)

Resources/doc/index.md
```diff
@@ -141,6 +141,22 @@ be omitted from the DSNs. Also make sure to use the sentinel port number
...
+``` yaml
+snc_redis:
+    clients:
+        default:
+            type: phpredis
+            alias: default
+            dsn: ["redis://localhost/0"]
+            options:
+                profile: 4.1
+                cluster: "phpredis-cluster"
```
B-Galati: This value can be anything but `null` or `false` right?

If so, I would suggest to validate this value to `true` or `false` or `null` in the extension, what do you think?

B-Galati:
@furester Thank you! I just set up dump a cluster locally (far from easy) I'll do further testing this week ;-)

furester:
@B-Galati i found some hint to setup a redis cluster in travis, but SncRedisBundle use ubuntu trusty (14.04 release), while redis cluster support were released on 1 Apr 2015. Do you think that it's possible to bump ubuntu version, or do you know how to use a newer version only for some tests cluster related?

B-Galati:
@furester It's super cool if you can achieve some integration/functional tests with a RedisCluster yet that will not block us to merge the PR.
What we need is at least testing the Extension to make sure that it does what we want now and in the future when configuring the bundle.
See `\Snc\RedisBundle\Tests\DependencyInjection\SncRedisExtensionTest` and `\Snc\RedisBundle\Tests\DependencyInjection\SncRedisExtensionEnvTest`

B-Galati:
PS: You can try to bump ubuntu version :+1:

Resources/doc/index.md
```diff
@@ -141,6 +141,22 @@ be omitted from the DSNs. Also make sure to use the sentinel port number
...
 
+A setup using `RedisCluster` from `phpredis`  could look like this:
+
+``` yaml
+snc_redis:
+    clients:
+        default:
+            type: phpredis
+            alias: default
+            dsn: ["redis://localhost/0"]
```
B-Galati: I saw that `RedisCluster` can accept multiple hosts that is called `seeds`. It's currently not possible and result in

```
In SncRedisExtension.php line 258:
                                                  
  [RuntimeException]                              
  Support for RedisArray is not yet implemented.  
```

This error message should be displayed only if `options.cluster` is not set. What do you think?

B-Galati:
`RedisFlushallCommand`, `RedisFlushdbCommand` are not working with the current implementation but IMHO it's fine and can be fixed later. I pushed a commit that disabled them with `RedisCluster`.

Factory/PhpredisClientFactory.php
```diff
@@ -28,40 +29,44 @@ public function __construct(RedisLogger $logger = null)
      * @param array  $options Options provided in bundle client config
      * @param string $alias   Connection alias provided in bundle client config
      *
-     * @return \Redis|Client
+     * @return \Redis|Client|\RedisCluster|ClientCluster
      * @throws InvalidConfigurationException
      */
     public function create($class, $dsn, $options, $alias)
```
B-Galati: It's important to update the tests of this class.

Resources/doc/index.md
furester: I agree with your suggestion, i'll add check asap.

furester:
@B-Galati little report: a couple of simple tests has been implemented, my first concern was to setup a redis whit cluster enable option and achieve this bumping ubuntu version and setting some `before_install` script in travis.

B-Galati:
@furester Awesome! Let me know when you'll need another round of review.

furester:
@B-Galati i did some refactoring in Extension and Factory, so we could manage \Redis and \RedisCluster client cleaner. 

Now i'm reading more about UnitTest, but i feel confident that you could review and eventually merge.

furester:
@B-Galati i did some refactoring in Extension and Factory, so we could manage \Redis and \RedisCluster client cleaner. But at the moment i didn't implement support for multiple hosts in cluster configuration. Do you think that's mandatory or we could add it later?

ps.: sorry for huge number of commits, i encountered a lot of trouble with array parameters in symfony integration.

Client/Phpredis/Client.php
```diff
@@ -224,23 +224,23 @@ public function setnx($key, $value)
     /**
      * {@inheritdoc}
      */
-    public function del($key1, $key2 = null, $key3 = null)
+    public function del($key, ...$other_keys)
     {
         return $this->call('del', func_get_args());
```
rpkamp: This actually changes the behaviour of this method when more than one parameter is passed.

When calling `del('key1', 'key2', 'key3);`

In the old situation this would translate to `$this->call('del', ['key1', 'key2', 'key3']);` but with the changed code it becomes `$this->call('del', ['key1', ['key2', 'key3']]);`

Instead `array_merge([$key], $other_keys)` should be used (or `[$key, ...$other_keys]`, but that is PHP7.4+, see https://wiki.php.net/rfc/spread_operator_for_array)

Client/Phpredis/ClientCluster.php
```diff
@@ -0,0 +1,1532 @@
...
+    /**
+     * Constructor.
+     *
+     * @param array       $parameters List of parameters (only `alias` key is handled)
+     * @param RedisLogger $logger     A RedisLogger instance
+     */
+    public function __construct(array $seeds = array(), array $parameters = array(), RedisLogger $logger = null)
+    {
+        $this->logger = $logger;
+        $this->alias = isset($parameters['alias']) ? $parameters['alias'] : '';
```
rpkamp: Null coalesce operator can be used `$this->alias = $parameters['alias'] ?? ''`

Client/Phpredis/ClientCluster.php
```diff
@@ -0,0 +1,1532 @@
...
+    {
+        return $this->call('setnx', func_get_args());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function del($key, ...$other_keys)
+    {
+        return $this->call('del', func_get_args());
```
rpkamp: Same problem as stated above, with more than 1 argument it will come `['key1', ['key2', 'key3']]`

rpkamp:
Other than those comments, looks good :+1:

Client/Phpredis/ClientCluster.php
B-Galati: fix

Client/Phpredis/Client.php
B-Galati: All fixed, thanks a lot for this great catch!

B-Galati:
@furester I squashed all commits, we are almost good to go. I'll ad support for multiple hosts soonish and then we can merge :-)

B-Galati:
:tada: Thanks @furester for this super great addition!

I'll open another PR for the multiple DSN support ;)

furester:
here i am, sorry .. busy days!
@B-Galati is this pr merged in master, so i can delete my fork, correct?

B-Galati:
@furester Correct!

firrae:
Will this be released for version 2.X @B-Galati? It seems with Predis going stale some of our code here can't use Predis cluster but a more direct implementation via PHPRedis should work. Sadly we are stuck on Symfony 2 though for the time being.

B-Galati:
@firrae Hi! No it will not but you can try to do a backport if you want and open a PR.

firrae:
@B-Galati thanks for the confirmation. We'll look into the possibility, but at this point I think we're leaning towards upgrading Symfony and then SnC to the 3.X version for this. But we'll see.

B-Galati:
That would be a better option yes :-)

PR #276, opened 2016-06-08
Fixes phpredis persistent connections when clients connect to different DBs

andreausu:
This fixes the corner case where clients connecting to different redis databases are used with persistent connection enabled by giving a unique ID to the persistent connections.

eg:

config:

```
snc_redis:
    clients:
        db1:
            type: phpredis
            alias: db1
            dsn: redis://localhost/1
            logging: %kernel.debug%
            options:
                connection_persistent: true

        db2:
            type: phpredis
            alias: db2
            dsn: redis://localhost/2
            logging: %kernel.debug%
            options:
                connection_persistent: true
```

code:

```
$db1 = $this->getContainer()->get('snc_redis.db1');
$db2 = $this->getContainer()->get('snc_redis.db2');
$db1->set('cachekey1', 'cachevalue');
$db2->set('importantkey2', 'importantvalue');
$db1->flushDb(); // db 2 will be flushed instead of db 1 
```

From phpredis documentation:

```
$redis->pconnect('127.0.0.1', 6379, 2.5, 'x'); // x is sent as persistent_id and would be another connection the the three before.
```

PR #462, opened 2018-10-13
[Fix][Predis] Fixes persistent connections when used many databases on the same instance

qRoC:
At the moment, the parameter value can only be boolean, so remote socket does not have a unique identifier. This is a potential error for those who use multiple databases on the same redis instance.

[Reference on predis code](https://github.com/nrk/predis/blob/v1.1/src/Connection/StreamConnection.php#L173)

Old PR(for phpredis):
https://github.com/snc/SncRedisBundle/pull/276
https://github.com/snc/SncRedisBundle/pull/308


curry684:
And tagged at https://github.com/snc/SncRedisBundle/releases/tag/2.1.7, thanks!

PR #469, opened 2018-11-02
Fix: No need to verify whether hard dependency is not null

localheinz:
This PR

* [x] removes what appears to be a condition that should not evaluate to `false`

### DIFF
--- a/DataCollector/RedisDataCollector.php
+++ b/DataCollector/RedisDataCollector.php
@@ -42,7 +42,7 @@ class RedisDataCollector extends DataCollector
     public function collect(Request $request, Response $response, \Exception $exception = null)
     {
         $this->data = array(
-            'commands' => null !== $this->logger ? $this->logger->getCommands() : array(),
+            'commands' => $this->logger->getCommands(),
         );
     }
 


DataCollector/RedisDataCollector.php
```diff
@@ -42,7 +42,7 @@ public function __construct(RedisLogger $logger)
     public function collect(Request $request, Response $response, \Exception $exception = null)
     {
         $this->data = array(
-            'commands' => null !== $this->logger ? $this->logger->getCommands() : array(),
```
localheinz: The `$logger` is a hard dependency, see 

https://github.com/snc/SncRedisBundle/blob/c5aaf5165301c0a5ad1f7af878512a7918ea2a96/DataCollector/RedisDataCollector.php#L34-L37

This condition only makes sense for scenarios where the constructor is not invoked:

* instantiated via reflection
* collector is overridden, but original constructor is not invoked

localheinz:
Build failure appears to be unrelated.

B-Galati:
Thanks @localheinz

curry684:
Thanks!

localheinz:
Thank you, @B-Galati, @curry684, and @OskarStark!

PR #470, opened 2018-11-02
Enhancement: Add note about usage with symfony/web-profiler-bundle

localheinz:
This PR

* [x] adds a note about usage with [`symfony/web-profiler-bundle`](https://github.com/symfony/web-profiler-bundle)

💁‍♂️ I have been scratching my head a bit, wondering why the Redis tab appears in the profiler, but no commands are listed, and from the documentation it's not quite obvious that it needs to be explicitly enabled.

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -380,6 +380,21 @@ snc_redis:
         ttl: 3600
 ```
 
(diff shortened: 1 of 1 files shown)


curry684:
Thanks!

PR #471, opened 2018-11-05
Supply name for new TreeBuilder instead of calling root method

rpkamp:
Creating a new TreeBuilder without root node information is
deprecated since Symfony 4.2.

See https://github.com/symfony/symfony/blob/master/UPGRADE-4.2.md#config

### DIFF
--- a/DependencyInjection/Configuration/Configuration.php
+++ b/DependencyInjection/Configuration/Configuration.php
@@ -11,6 +11,7 @@
 
 namespace Snc\RedisBundle\DependencyInjection\Configuration;
 
+use function method_exists;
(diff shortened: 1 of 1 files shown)


DemigodCode:
Hey, when will a new tag with this change will be created?

curry684:
As it was merged into `master` not until all other changes blocking a 3.x release have been handled.

curry684:
Mmm good point I'll cherrypick it.

curry684:
Done in https://github.com/snc/SncRedisBundle/commit/4ddb2494d64012b5f75aa7ffc3b34ee7346a16ba, did have to remove the `use function` line as it's PHP 5.6+.

PR #473, opened 2018-11-16
Fix: Remove useless else

localheinz:
This PR

* removes a useless `else`


localheinz:
Thank you, @B-Galati and @curry684!

PR #477, opened 2018-11-29
Fix: Make script step explicit

localheinz:
This PR

* [x] makes the `script` step explicit

💁‍♂️ For reference, see https://docs.travis-ci.com/user/languages/php#default-build-script.


localheinz:
Thank you, @curry684!

PR #482, opened 2018-12-11
Update Travis matrix

rvanlaak:
As PHP 7.0 is EOL and PHP 7.1 unsupported 

http://php.net/supported-versions.php

Hopefully this also fixes the failing `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,12 @@ services:
 
 matrix:
   include:
-    - php: 7.0
+    - php: 7.2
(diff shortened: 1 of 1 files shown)


curry684:
The fails in Master are something we can tackle once we have merged the 2.1 changes into it.

I don't intend to drop support for PHP versions the day their creators drop support for it, as it would alienate tens of thousands of developers out there using this on servers that cannot or will not be upgraded beyond 7.0 or 7.1, some even for good defendable reasons.

So I'm closing this as I hold the opinion that library developers should not drop support for older upstream/PHP versions until they need to or consciously choose to upgrade for newer language features. Right now neither applies.

PR #489, opened 2019-01-11
set logger on the connection wrapper only

xabbuh:
The `Predis\Connection\Factory` class does not have a `setLogger()` method.

### DIFF
--- a/Client/Predis/Connection/ConnectionFactory.php
+++ b/Client/Predis/Connection/ConnectionFactory.php
@@ -65,8 +65,8 @@ class ConnectionFactory extends Factory
             if (null !== $this->wrapper) {
                 $wrapper = $this->wrapper;
                 $connection = new $wrapper($connection);
+                $connection->setLogger($this->logger);
             }
-            $connection->setLogger($this->logger);
         }
 
         return $connection;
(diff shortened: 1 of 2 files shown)


B-Galati:
@xabbuh Thx! Could you target branch 2.1 as it's a bugfix please?

xabbuh:
@B-Galati sure, rebase done

micayael:
Hi. Is there any update on this?

B-Galati:
@micayael Please try the fix and let us know if it works for you

micayael:
> @micayael Please try the fix and let us know if it works for you

Yes. It works

B-Galati:
I am not a fan of this solution as it would mean that logging could be activated but not working without knowing it. I would prefer something that throw a LogicException. What do you think all?

xabbuh:
I am unfortunately not able to reproduce the initial bug report. I also think that such a solution could be a bit harder if the logging config and the connection class to be used cannot be evaluated at the same time.

I personally would merge this PR first to mitigate the reported issue and then think about how the issue you raised could be handled in a way that provides better feedback.

B-Galati:
@xabbuh Could you rebase your work please? CI should be fixed after that.

xabbuh:
Hm, one of the jobs is still failing. I am going to check what's causing this.

B-Galati:
@xabbuh Looks weird, let me know if you need some help

B-Galati:
The test is skipped on `master` branch as it is by nature unreliable so I guess we could do the same here.
Could you do so? I'll merge it then.

xabbuh:
@B-Galati done

PR #492, opened 2019-01-17
No memory limit for composer in TravisCI

B-Galati:


B-Galati:
Merged, let me know for anything

PR #499, opened 2019-02-19
fix #498 - Add read_timeout connection option to phpredis

B-Galati:
fix #498

PR #500, opened 2019-02-19
Fixed read_write_timeout ignored when using phpredis

agus-k:
Patch to fix read_write_timeout option being ignored when using phpredis


B-Galati:
Thanks! Could you review #499. It's another solution that does not hardcode reconnect delay.

agus-k:
More proper fix in #499

PR #501, opened 2019-02-21
fix #383 - deprecates redis service alias in favor of 'snc_redis.{alias}'

### LINKED ISSUE #383: Predis vs Phpredis container definitions
Is there a reason that a Predis client gets the [service definition](https://github.com/snc/SncRedisBundle/blob/master/DependencyInjection/SncRedisExtension.php#L205) name `snc_redis.CLIENT_ALIAS` but a Phpredis client gets the [service definition](https://github.com/snc/SncRedisBundle/blob/master/DependencyInjection/SncRedisExtension.php#L304) name `snc_redis.phpredis.CLIENT_ALIAS` and an alias to `snc_redis.CLIENT_ALIAS`

I only noticed because we have a couple of instances (using the oro platform, part of the cache dumping) where services are requested by `getDefinition` [method](https://github.com/symfony/dependency-injection/blob/master/ContainerBuilder.php#L967) rather than `findDefinition` [method](https://github.com/symfony/dependency-injection/blob/master/ContainerBuilder.php#L989) and so, when using the `snc_redis.CLIENT_ALIAS` key, Predis clients don't throw an error because they have an actual definition but Phpredis client definitions cannot be found as `getDefinition` does not look for aliases.

Would be interested to hear your thoughts

B-Galati:


B-Galati:
@curry684 Does it look ok to you?

curry684:
Doesn't this implementation also trigger the deprecation warning on every container compile, not only if you're actually using it?

(dead tired and friday night so might be missing something)

B-Galati:
@curry684 Yes it will, I did not find another to tell the user that this alias is deprecated sadly :/

B-Galati:
@curry684 Do you have any more concerns ?
Do you think we should remove the deprecation for all container compile and just having it documented in UPDATE.md ?

curry684:
Think it's more elegant to just consider it a breaking change in documentation. I consider a deprecation warning something that I can fix, and once my code has 0 deprecation warnings I can safely start upgrade to next major.

B-Galati:
@curry684 Thanks! It makes a lot of sense. I just removed them.

B-Galati:
@curry684 Let me know for anything else or if I can merge it ;-)

PR #502, opened 2019-02-21
fix #383 - remove redis service alias in favor of 'snc_redis.{alias}'

### LINKED ISSUE #383: Predis vs Phpredis container definitions

B-Galati:
Should be rebased once #501 is on master


DependencyInjection/SncRedisExtension.php
```diff
@@ -197,7 +197,7 @@ protected function loadPredisClient(array $client, ContainerBuilder $container)
         $optionDef->addArgument($client['options']);
         $container->setDefinition($optionId, $optionDef);
         $clientDef = new Definition($container->getParameter('snc_redis.client.class'));
-        $clientDef->setPublic(false);
+        $clientDef->setPublic(true);
```
B-Galati: I don't see any good reason to keep these aliases private while they are meant to be used by user of the bundle.
IMHO, enforcing the usage of the service locator would be something cumbersome.
Should we drop the service locator feature?
Perhaps I am missing something here.

curry684: Services should, since Symfony 3.4, be private "unless very good reason to make public". The reason for this is that the container compilation performs tree shaking steps that can never predict when a public service is being used. So by requiring explicit injection, either through autowiring, config, service locators or service subscribers, the container can be compiled more efficiently, smaller and faster.

So given that we're doing a new major now I would definitely "take the break" and make them all private by default.

curry684: https://symfony.com/blog/new-in-symfony-3-4-services-are-private-by-default

B-Galati: Thanks! Shame on me! How could I forget about all of these!
I'll revert this change.

B-Galati:
LGTM, let me know for anything

PR #505, opened 2019-03-13
Leverage "options.parameters" config in `PhpredisClientFactory::create()`

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |no    |
|New feature? |no    |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|n/a   |
|License      |MIT   |
|Doc PR       |n/a   |

### DIFF
--- a/Factory/PhpredisClientFactory.php
+++ b/Factory/PhpredisClientFactory.php
@@ -68,10 +68,14 @@ class PhpredisClientFactory
 
         if (null !== $parsedDsn->getPassword()) {
             $client->auth($parsedDsn->getPassword());
+        } elseif (isset($options['parameters']['password'])) {
+            $client->auth($options['parameters']['password']);
         }
 
         if (null !== $parsedDsn->getDatabase()) {
             $client->select($parsedDsn->getDatabase());
+        } elseif (isset($options['parameters']['database'])) {
+            $client->select($options['parameters']['database']);
         }
 
         if (isset($options['read_write_timeout'])) {
--- a/Tests/DependencyInjection/SncRedisExtensionTest.php
+++ b/Tests/DependencyInjection/SncRedisExtensionTest.php
@@ -368,6 +368,46 @@ class SncRedisExtensionTest extends TestCase
         $this->assertEquals(array('snc_redis.default' => array(array('alias' => 'default'))), $container->findTaggedServiceIds('snc_redis.client'));
     }
 
+    /**
+     * Test provided options are respected
+     */
+    public function testPhpRedisParameters()
+    {
+        $extension = new SncRedisExtension();
+        $config = $this->parseYaml($this->getPhpRedisYamlConfigWithParameters());
+        $extension->load(array($config), $container = $this->getContainer());
+
+        $defaultParameters = $container->getDefinition('snc_redis.phpredis.default');
+
+        $this->assertSame(1, $defaultParameters->getArgument(2)['parameters']['database']);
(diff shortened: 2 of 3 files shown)


phansys:
Without these changes, parameters provided at `snc_redis.clients.{name}.options.parameters` are completely ignored within "phpredis". If this was the desired behavior, I think the current branch is right; otherwise, this must be considered a fix and the base should be updated to target `2.1`.
Thank you in advance.

B-Galati:
Thanks @phansys! Could you target `2.1` please?

Factory/PhpredisClientFactory.php
```diff
@@ -68,10 +68,14 @@ public function create($class, $dsn, $options, $alias)
 
         if (null !== $parsedDsn->getPassword()) {
             $client->auth($parsedDsn->getPassword());
+        } elseif (isset($options['parameters']['password'])) {
```
B-Galati: Could you add unit tests for these cases please?

phansys: I have added tests: https://github.com/snc/SncRedisBundle/pull/505/files#diff-faa94e9892a5c176d4921a4a202f29f5R677. Are not they enough? Please, let me know.

phansys:
Updated target to `2.1`.

phansys:
I had to use the full service definition in tests ("snc_redis.phpredis.default") instead of the short alias ("snc_redis.default" ) in tests since it seems to be unavailable in tests for phpredis. Do you have any clue about this?
https://github.com/snc/SncRedisBundle/blob/3f64274b5ca32f93706b19aac970237ba0e949d3/DependencyInjection/SncRedisExtension.php#L300

phansys:
Given my last comment, I think this PR should wait for #506 in order to fix the service definition.

Factory/PhpredisClientFactory.php
B-Galati: Yes and that is great!
I think we also need to update SncRedisBundle/Factory/PhpredisClientFactoryTest.php.

B-Galati:
It's ok if you just use `snc_redis.phpredis.default` service id for tests, this will be fixed in future 3.0.0 where only short id will be kept.

Factory/PhpredisClientFactory.php
phansys: That's right. Updated.

phansys:
This PR should wait #507 to fix the usage of phpredis through `Client` class.

B-Galati:
I don't get why we did not have this error before :thinking:

B-Galati:
I get it, the test was not executed at all before :sweat_smile: 
Thank you for moving it :+1:

phansys:
@B-Galati, in order to get this PR merged and separated concerns, I've restored `PhpredisClientFactoryTest` to its original path (keeping the changes related to this PR). I've created #508 in order to address the location change.
Thank you in advance.

B-Galati:
@phansys Please keep everything in this one as it is related in some way IMHO.

Also tests that depends on a specific phpredis version are usually handled like so:

https://github.com/snc/SncRedisBundle/blob/521d237f93ec4b5ec089496924faf061130097a0/Tests/Client/Phpredis/ClientTest.php#L18-L22

No need to bother with redis4 group or something like this.

phansys:
Done @B-Galati.

B-Galati:
@phansys Awesome! Thanks a lot!

Issue #442, opened 2018-07-20
2.1.5 (lazy loading) somehow breaks set method


sawmurai:
Hey guys,

the lazy-loading introduced with the latest patch somehow seems to break the `set` method. 

When calling it, the following error pops up:

```
Redis::set() expects at most 3 parameters, 4 given
```

although it gets called with 3 parameters (in a controller):

```php
    private function setRefreshTokenCachedResponse(Request $request, Response $response)
    {
        $redis = $this->container->get('snc_redis.session');
        $redis->set(
            $request->get('refresh_token'),
            $response->getContent(),
            self::REFRESH_TOKEN_RESPONSE_CACHE_LIFETIME
        );
    }
```

dkarlovi:
Talking about it with @ocramius here: https://twitter.com/dkarlovi/status/1020271285320941568

Ocramius:
Please also show the generated code for the proxy.

sawmurai:
[Redis_ca5fc0f.php.txt](https://github.com/snc/SncRedisBundle/files/2213796/Redis_ca5fc0f.php.txt)

You mean this?

dkarlovi:
https://gist.github.com/dkarlovi/cd0174593bd6b0873afc803bd5a43978

dkarlovi:
So, as [noted](https://github.com/phpredis/phpredis#set):

> Timeout or Options Array (optional). If you pass an integer, phpredis will redirect to SETEX, and will try to use Redis >= 2.6.12 extended options if you pass an array with valid values

It should be one of, they're both included for some reason.

Ocramius:
Yep, exactly that. 

And holy crap this class has an atrocious API: just how many responsibilities are crammed together here?!

Besides that, as you can see, `set()` is generated with 4 parameters, of which 2 optional.

To keep in mind when going further here:

 * ext-reflection is seeing 4 parameters here
 * default parameter values of methods defined in core or in an extension cannot be reflected, hence `NULL` as default value.

curry684:
Yep, phpredis is a mess, it's essentially just [all Redis commands](https://redis.io/commands) thrown together in a big bucket class. Considering that's how Redis itself works it's not even unreasonable but it makes wrapping hell.

Marking this issue as a not-our-problem-but-we-need-to-keep-it-in-mind issue.

dkarlovi:
Opened the issue: https://github.com/phpredis/phpredis/issues/1374

curry684:
[...]

PR #509, opened 2019-03-16
[WIP] Build custom Redis client based on the installed extension

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |no    |
|New feature? |yes   |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|#399  |
|License      |MIT   |
|Doc PR       |n/a   |


phansys:
I'm trying to use the reflection API in order to introspect the `\Redis` class, but it seems to fail to get some method arguments. By instance, with `pconnect()` it doesn't find any argument. Do you know how could I get a valid representation of the class provided by the extension?
Thank you in advance.

B-Galati:
See #442 if you want more details ;-)

.travis.yml
```diff
@@ -56,4 +59,6 @@ before_install:
 
 install: composer update
 
-before_script: mkdir Snc && ln -s ../ Snc/RedisBundle
+before_script:
+  - mkdir Snc && ln -s ../ Snc/RedisBundle
+  - php clientbuilder/build_redis_client.php
```
B-Galati: I am not sure why it is necessary to have that in the CI.

IMHO, the class should be generated at cache warmup then dump into var/cache/snc-redis-bundle and then loaded thanks to a custom autoloader. It's basically what doctrine-bundle is doing for entities.

What do you think?

phansys: I think that could work only when using DI, but be aware that `PhpredisClientFactory` doesn't depend on the container.
Also, in my first approach I was thinking to check the generated file in each build in order to detect any change based on the installed Redis version, but I think it doesn't have sense now that the installed versions are set explicitly at Travis config.

phansys:
Thank you for pointing me in the right direction. I've opened phpredis/phpredis#1526 for the described problem.

B-Galati:
@phansys Given what was said in the issue, I think we could only support phpredis4 for this feature and have a fallback compatible with phpredis3 only. That's quite easy to do since v3 will not change anymore.
We could provide another one for phpredis2 but it's not important IMHO as it's quite an old version.

phansys:
Sure. I'll try to work on the proposed fix today. Thank you.

phansys:
Methods like `echo()` and `eval()` are present at Redis 3.1.6, but these are not allowed by PHP < 7.0. Should we remove them from the custom client based on that version or should we restrict the usage of the client only for PHP >= 7?
For BC reasons, I think the first option is more suitable for this case, but I'd like to hear your opinion.

B-Galati:
> By the other hand, Redis `4.3.0` seems to be not compatible with `4.2.0` ([ref](https://travis-ci.org/snc/SncRedisBundle/jobs/508445757)):
> 
> ```
> Declaration of Snc\RedisBundle\Client\Phpredis\Client4_3_0::psubscribe(array $patterns, $callback) should be compatible with Redis::psubscribe(array $patterns)
> ```
> 
> Should we ship a client by minor version?

For phpredis4 we could only generate the client class on cache warmup so that it would always fit with the installed phpredis version.

We would have static client class only for older phpredis version, (e.g.) v2 and v3.


> Methods like `echo()` and `eval()` are present at Redis 3.1.6, but these are not allowed by PHP < 7.0. Should we remove them from the custom client based on that version or should we restrict the usage of the client only for PHP >= 7?
> For BC reasons, I think the first option is more suitable for this case, but I'd like to hear your opinion.

For v3, the bundle only accepts PHP 7 so let's say we add this feature for V3 only. What do you think?
Which should be released in less than 2 months (I hope).

phansys:
I think your proposal is doable. Regarding the release of `v3.0.0`, should this PR target `master`?

B-Galati:
Yes `master` for upcoming V3.

B-Galati:
Should we close this PR ?

phansys:
>Should we close this PR ?
Yes, closing since this fix will not be accepted as is for `2.x` branch.

PR #511, opened 2019-03-20
[3.x] Build custom Redis client based on the installed extension

### LINKED ISSUE #399: (bc) problem with 4.0 redis

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |no    |
|New feature? |no    |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|#399  |
|License      |MIT   |
|Doc PR       |n/a   |

Fixes #399.

This PR reverts the changes made at #405 and #487.


phansys:
@B-Galati, how do you propose to load the cache generated class? Currently, I'm generating the file as `snc_redis_client.php` inside the cache dir:
https://github.com/snc/SncRedisBundle/blob/e1da2fd9446a11fa2c667244b934bd19fdc61996/Client/Phpredis/ClientCacheWarmer.php#L41

B-Galati:
Good question.
Could be done with a custom autoloader like doctrine is doing.
You will need to double check because I am not 💯 sure.

Tests/Factory/PhpredisClientFactoryTest.php
```diff
@@ -1,44 +1,47 @@
...
+        if (!class_exists(\Redis::class)) {
             $this->markTestSkipped(sprintf('The %s requires phpredis extension.', __CLASS__));
         }
 
         if (!@fsockopen('127.0.0.1', 6379)) {
             $this->markTestSkipped(sprintf('The %s requires a redis instance listening on 127.0.0.1:6379.', __CLASS__));
         }
 
-        $this->logger = $this->getMockBuilder('Symfony\Component\HttpKernel\Log\LoggerInterface')->getMock();
+        $this->logger = $this->getMockBuilder(LoggerInterface::class)->getMock();
```
B-Galati: That's not the same class as before.

phansys: It isn't because the previous class was removed: https://github.com/symfony/http-kernel/commit/897836f4af4ecd620804c64e76949e1383ce4330

B-Galati: Indeed! Thanks!

I am wondering how it could have worked before :thinking:

B-Galati:
@curry684 Any advice about the implementation? It looks like you thought about it a lot in the past.

Client/Phpredis/ClientCacheWarmer.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    private $clientBuilder;
+
+    public function __construct(ClientBuilder $clientBuilder)
+    {
+        $this->clientBuilder = $clientBuilder;
+    }
+    public function isOptional()
+    {
+        return class_exists(Client::class);
```
B-Galati: Client::class is not imported. Is it normal?

FYI the warmup is not needed if no phpredis clients are used.

SncRedisBundle.php
```diff
@@ -23,6 +24,16 @@
  */
 class SncRedisBundle extends Bundle
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function boot()
+    {
+        if (!class_exists(Client::class) && file_exists($file = $this->container->getParameter('kernel.cache_dir').'/snc_phpredis_client.php')) {
+            require $file;
```
B-Galati: Good idea :+1:

SncRedisBundle.php
```diff
@@ -23,6 +24,16 @@
  */
 class SncRedisBundle extends Bundle
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function boot()
+    {
+        if (!class_exists(Client::class) && file_exists($file = $this->container->getParameter('kernel.cache_dir').'/snc_phpredis_client.php')) {
```
B-Galati: What do you think about throwing an exception if a phpredis client is defined but the file is not present in the cache ?

The mechanism would be more robust and we could identify mistakes easily.

B-Galati:
To detect the phpredis usage, it  could be a parameter set by the extension as soon as a phpredis client is declared like here https://github.com/snc/SncRedisBundle/blob/521d237f93ec4b5ec089496924faf061130097a0/DependencyInjection/SncRedisExtension.php#L247-L295

Client/Phpredis/ClientCacheWarmer.php
B-Galati: Testing for `\Redis::class` existence would be enough if the `boot()`  method of the bundle throw an exception like I proposed.

phansys:
The suggested changes were addressed, I think. Now I'm facing 2 new issues:
* The cache seems to be not cleared after the test `CustomPhpredisClientTest`, causing other tests to fail;
* Installing Redis 4.3.0 is causing [segfault](https://travis-ci.org/snc/SncRedisBundle/jobs/509713216) at Travis builds.

B-Galati:
@phansys Looks like you fix them! Nice! What were the issues?

phansys:
These are the changes I've done to solve the described issues:
* Check against the class existence regardless the cache file exists or not;
* Bundle the Redis library before installing it (`pecl bundle redis`), avoiding to force "yes" for all the prompts triggered when running `pecl install redis`. Maybe some of the flags was causing the segfault.
Now I must try to solve the issues with the client for `3.1.6` version.

phansys:
I think everything is working now @B-Galati. Thank you for your support!

B-Galati:
@phansys my pleasure!

I'll play with it and give you some feedback!
This feature is gonna be awesome :+1: 

@curry684 Would you have time to check it out too? The more we review the better ;-)

B-Galati:
With these changes we could try to get rid of this limitation https://github.com/snc/SncRedisBundle/blob/86b5d1e4a82701bb172800106bb1c719a03bb1bd/DependencyInjection/SncRedisExtension.php#L283-L288

phansys:
> With these changes we could try to get rid of this limitation

Done.

B-Galati:
@phansys Hi! I am currently checking @furester work and I saw he used `...$args` to deal with method compatibility issues. Now that we support only php 7.1+ it would be a great solution and avoid the burden of maintaining the dynamic generation of proxy class.

I tested it like this https://3v4l.org/9OMNC

Regarding https://github.com/snc/SncRedisBundle/pull/511#issuecomment-476856537 it's not gonna work so would be nice to revert that change sorry :/.

B-Galati:
@phansys Nevermind about `...$args` stuff I forgot to implements a class :man_facepalming: it does not work.

phansys:
>@phansys Nevermind about ...$args stuff I forgot to implements a class man_facepalming it does not work.

I rebased the PR and added support for builds with Redis 5.0.2.

B-Galati:
@phansys What about using `symfony/proxy-manager-bridge` instead of creating our own logic?

phansys:
> @phansys What about using `symfony/proxy-manager-bridge` instead of creating our own logic?

Since we need to intercept each call to Redis' methods, I think we should use [`AccessInterceptorValueHolderFactory`](https://github.com/Ocramius/ProxyManager/blob/master/docs/access-interceptor-value-holder.md#access-interceptor-value-holder-proxy); but I don't know how to achieve that with `symfony/proxy-manager-bridge`.
I've never developed a proxy using this library. Could you please point me in the right direction?

Thank you.

B-Galati:
@phansys Sorry i never used it either :/

But I see 2 benefits, the underlying lib (`Ocramius/ProxyManager`) is battle tested and it will make less complicated code to maintain. 

(I could be missing some points that would make the usage of this lib impossible or hard)

qharnay:
Hi, it's been a while since last comment ...  Is there any news of acceptance for this PR ?

Cheers

B-Galati:
Hey, I am not sure about the maintenance cost of this solution.

I am currently not having time to look at such a big feature and maintain it after the release.

What's your opinion @curry684?

curry684:
> I am currently not having time to look at such a big feature and maintain it after the release.

Same issue here... I think the intended direction is the right way (as discussed long ago) but indeed it's far from trivial to both develop and maintain, and depending on phpredis itself it may either be fragile or more resilient than what we have now.

*If* we go forward on this I would most certainly prefer using the proxy manager as that's indeed battle tested and well maintained.

ostrolucky:
Superseded by https://github.com/snc/SncRedisBundle/pull/636

PR #515, opened 2019-04-08
Wrap Parameters in array when using replication - fix #381

### LINKED ISSUE #381: Single \Predis\Connection\Parameters instance should be wrapped in array when using sentinel replication.
Config below results in  `ERR unknown command 'SET'` when trying to set session data. Parsed DSN is passed as an instance of `\Predis\Connection\Parameters` to `\Predis\Client` which results in sending SET/GET to Redis Sentinel instead of real Redis node.

```
snc_redis:
    clients:
        session:
            type: predis
            alias: session
            dsn:
                - redis://redis-sentinel:26379                 
            options:
                replication: sentinel
                service: master
```

duxet:


B-Galati:
@duxet Thank you! Could you target 2.1 branch please?

PR #517, opened 2019-05-15
Adding metrics to Profiler page

stefantalen:
I was working on a project that did not have the toolbar available in responses (JSON) so I was debugging using the page within the profiler. I found some of the metrics from the toolbar quite useful so hence this PR.

### DIFF
--- a/Resources/views/Collector/redis.html.twig
+++ b/Resources/views/Collector/redis.html.twig
@@ -113,12 +113,10 @@
                 <span class="label">Query time</span>
             </div>
 
-            {% if collector.erroredCommandsCount > 0 %}
-                <div class="metric highlight">
-                    <span class="value error">{{ collector.erroredCommandsCount }}</span>
-                    <span class="label">Failed Queries</span>
-                </div>
-            {% endif %}
+            <div class="metric highlight">
+                <span class="value{% if collector.erroredCommandsCount > 0 %} error{% endif %}">{{ collector.erroredCommandsCount }}</span>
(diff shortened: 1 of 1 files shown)


B-Galati:
@stefantalen Thank you!

I'll look into it asap

stefantalen:
ping @B-Galati

B-Galati:
@stefantalen I think I will be able to check it out in July ;-)

Resources/views/Collector/redis.html.twig
```diff
@@ -102,6 +102,25 @@
...
+                <span class="value">{{ collector.commandcount }}</span>
+                <span class="label">Queries</span>
+            </div>
+
+            <div class="metric">
+                <span class="value">{{ '%0.2f'|format(collector.time) }} <span class="unit">ms</span></span>
+                <span class="label">Query time</span>
+            </div>
+
+            {% if collector.erroredCommandsCount > 0 %}
```
IonBazan: I think we should still show **0 Failed Queries**

stefantalen: Good suggestion 👍

I've updated the view

B-Galati:
Nice job thanks!

image

stefantalen:
Thanks for merging 👍 , will it be available in the next release (2.1.12)?

B-Galati:
@stefantalen No you'll need to update to the next major version.

The upgrade should not be difficult check our UPGRADE.md file.

PR #524, opened 2019-07-30
fix #523: use a new service id insteal of an alias

### LINKED ISSUE #523: Error on deprecation message
Hello,

I updated my project with the version 2.1.10, and I've seen that there is a deprecation on the bundle service name.
But it seems that the message in the logs is incorrect, because the same service name appears twice :

`[2019-07-30 10:28:41] php.INFO: "snc_redis.front" service is deprecated since 2.1.10, to be removed in 3.0. Please use "snc_redis.front" instead.`

The error seems to come from SncRedisExtension.php in line 307.

Am I the only one having this error ?

B-Galati:


HitoTech:
I tried it on my environment and it's working!

But the tests seems to be broken on Travis

B-Galati:
I think I will remove the deprecation message because having 2 different services could lead to 2 connection to Redis for the same service and it could be harmful.

B-Galati:
What do you think @curry684?

curry684:
Yeah seems legit, and not like we have Symfony-level obligations to deprecate every feature with full B/C promise. It's just a thing that'll break right away in a major upgrade anyway.

B-Galati:
Fixing it using alias deprecation of SF 4.3, easy and better than nothing.

image

PR #526, opened 2019-08-15
Fix C&P error in Phpredis Client Proxy Class

deegital:
$key is not provided for scan() - seems like due to some C&P action the method body of some of the other scan methods was used.


B-Galati:
thank you!

PR #528, opened 2019-08-27
Autoconfigure RedisBaseCommand to add snc_redis.command tag

maxhelias:
Autoconfigure instances of `Snc\RedisBundle\Command\RedisBaseCommand` to add `snc_redis.command` tag

### DIFF
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -11,6 +11,7 @@
 
 namespace Snc\RedisBundle\DependencyInjection;
 
+use Snc\RedisBundle\Command\RedisBaseCommand;
 use Snc\RedisBundle\DependencyInjection\Configuration\Configuration;
 use Snc\RedisBundle\DependencyInjection\Configuration\RedisDsn;
 use Snc\RedisBundle\DependencyInjection\Configuration\RedisEnvDsn;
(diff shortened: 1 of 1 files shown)


B-Galati:
Should we remove the tag `snc_redis.command` from `redis.xml` too?

maxhelias:
I really do not know why the configuration that is retrieved from the redis.xml file is not update. But when I test my custom redis commands of my project it works well.

There would not be a particularity on the conf files ? I see on the Symfony project, the command tag is redefined.

rpkamp:
FYI the entire tag will be obsolete if https://github.com/snc/SncRedisBundle/pull/533 is merged.

B-Galati:
Thanks @maxhelias

B-Galati:
I don't get why you're saying that command tag would be obsolete?

It's still needed to inject the service locator no?

B-Galati:
> I really do not know why the configuration that is retrieved from the redis.xml file is not update. But when I test my custom redis commands of my project it works well i see the tag and i don't see the tag on `flushhdb` command if i remove it on the config file.
> 
> There would not be a particularity on the conf files ? I see on the Symfony project, the command tag is redefined.

@maxhelias I tried and don't get that behavior either so :+1:

B-Galati:
Any hint @nicolas-grekas ?

rpkamp:
If we'd merged #533 it would be obsolete. But we didn't, because #533 was a bad idea.

So, it's not obsolete, this PR is a good idea, and it should be merged :slightly_smiling_face:

B-Galati:
All clear now thanks!

PR #532, opened 2019-09-04
No longer use curly brackets for substring

rpkamp:
It has been deprecated since PHP 7.4

See https://wiki.php.net/rfc/deprecate_curly_braces_array_access


B-Galati:
Thank you!

PR #533, opened 2019-09-14
Inject service locator into console commands

rpkamp:


maxhelias:
It's a good move but i think we can't remove the tag `snc_redis.command` like that. We need to keep running for command without inheriting the class `RedisBaseCommand.php` and warn the user of the change to be made.
WDYT ?

rpkamp:
I had not considered that, but yes that does make sense.

I'll update my PR.

rpkamp:
So I set out to implement what you suggested and have come to the conclusion that it is near impossible to do it.

The main issue is with autowiring of any custom commands people may have created in userland. Ideally they should be marked as child definitions of the `RedisBaseCommand` class, but marking a parent using autoconfiguration is not possible.
Moreover, since the `RedisBaseCommand` is abstract, any class that extends it is free to change the constuctor as they please (as per default PHP behaviour). So, it is not trivial to figure out where the `ServiceLocator` should be injected. At least without relying on reflection, which I'd rather not use.

So all in all I think it would be better to keep the `snc_redis.command` tag and the `setClientLocator` method on the `RedisBaseCommand` and only add a `snc_redis.client_locator` service in the container which is configured in the `ClientLocatorPass` as already proposed in this PR.
That would already be an improvement over the current implementation, as it's simpler and more explicit.

Any thoughts?

maxhelias:
With the configuration of this PR, the `snc_redis.command` tag will no longer be useful event at `RedisBaseCommand`. So by definition all the remaining tags will be linked to a custom implementation of the user without inherit this class.
From there we can already send a warning to the user.
After that we can check if the custom command of the user has the method `setClientLocator` where we can pass him the `ServiceLocator` and if not we can do nothing else.

I think we can do it unless I miss something ?

rpkamp:
I'll create an alternative PR with what I'm describing above later. Maybe that makes it more clear.

rpkamp:
See https://github.com/snc/SncRedisBundle/pull/535/files

rpkamp:
Closing this in favour of https://github.com/snc/SncRedisBundle/pull/535

PR #538, opened 2019-10-08
Fix check for connection_persistent with phpredis factory

### LINKED ISSUE #537: Phpredis factory adds wrong parameter for non-persistent connections
The `PhpredisClientFactory` does an incorrect check at line https://github.com/snc/SncRedisBundle/blob/86b5d1e4a82701bb172800106bb1c719a03bb1bd/Factory/PhpredisClientFactory.php#L58

If the value is `false`, the condition is true, see this screenshot, and adds an extra parameter to the call, the "PersistentId", which does not make sense.

Screenshot 2019-10-08 at 17 00 02

starred-gijs:
Fixes #537 

This condition is now exactly the same as the one to check for connection method, see https://github.com/snc/SncRedisBundle/blob/86b5d1e4a82701bb172800106bb1c719a03bb1bd/Factory/PhpredisClientFactory.php#L62

### DIFF
--- a/Factory/PhpredisClientFactory.php
+++ b/Factory/PhpredisClientFactory.php
@@ -55,7 +55,7 @@ class PhpredisClientFactory
             $connectParameters[] = null;
         }
 
-        if (isset($options['connection_persistent'])) {
+        if (!empty($options['connection_persistent'])) {
             $connectParameters[] = $parsedDsn->getPersistentId();
         }
 


B-Galati:
@starred-gijs Could you add some tests to PhpredisClientFactoryTest please?

starred-gijs:
Ah yes will do that, did not realise there was a testcase for this class. Will do it shortly

starred-gijs:
Hmmz now that I have seen the testcase, Im not sure how to write a test for this, any pointers?

B-Galati:
@starred-gijs You will need to check what's the value of `$client->getPersistentID()` in current test cases.

B-Galati:
Also `'connection_persistent' => true,` looks wrong as it should be an integer.

B-Galati:
And my bad it should be string according to phpredis doc https://github.com/phpredis/phpredis#pconnect-popen

B-Galati:
Merging it. I will add some tests later before the release tomorrow morning.

Thanks a lot @starred-gijs!

PR #552, opened 2019-12-09
Fix php 7.4 deprecation

mops1k:


B-Galati:
CI fail is unrelated

B-Galati:
thanks @mops1k

Issue #550, opened 2019-12-02
Setting cluster option to false is not recognise.


Webonaute:
Using this config does not allow to set cluster on/off.

`snc_redis:
  clients:
    default:
      type: phpredis
      alias: default
      dsn:
        - '%cache_host%'
      options:
        prefix: '%api_version%_'
        cluster: '%cache_dns_is_cluster%'`

When looking at the code, it assume that if isset cluster (not null), than its true.
SncRedisExtension class near line 262 and 214.

B-Galati:
You are right, would you mind sending a PR?

PR #553, opened 2019-12-09
Allow cluster option to be set to false.

Webonaute:
https://github.com/snc/SncRedisBundle/issues/550

### DIFF
--- a/DependencyInjection/SncRedisExtension.php
+++ b/DependencyInjection/SncRedisExtension.php
@@ -150,9 +150,7 @@ class SncRedisExtension extends Extension
      */
     protected function loadPredisClient(array $client, ContainerBuilder $container)
     {
-        if (null === $client['options']['cluster']) {
-            unset($client['options']['cluster']);
-        } else {
+        if (true === $client['options']['cluster']) {
             unset($client['options']['replication']);
         }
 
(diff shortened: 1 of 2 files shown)


DependencyInjection/SncRedisExtension.php
```diff
@@ -150,9 +150,7 @@ protected function loadClient(array $client, ContainerBuilder $container)
      */
     protected function loadPredisClient(array $client, ContainerBuilder $container)
     {
-        if (null === $client['options']['cluster']) {
-            unset($client['options']['cluster']);
-        } else {
+        if (true === $client['options']['cluster']) {
```
B-Galati: it should test if the value is different than `false` instead because Predis is using string to configure cluster

see https://github.com/nrk/predis#client-configuration

DependencyInjection/SncRedisExtension.php
```diff
@@ -150,9 +150,7 @@ protected function loadClient(array $client, ContainerBuilder $container)
      */
     protected function loadPredisClient(array $client, ContainerBuilder $container)
     {
-        if (null === $client['options']['cluster']) {
-            unset($client['options']['cluster']);
```
B-Galati: Is this change tested? Because I am not sure if it will break Predis behavior or not.

Issue #513, opened 2019-03-25
How to use the database parameter


peter-gribanov:
```yml
snc_redis:
    clients:
        default:
            type: predis
            alias: default
            dsn: '%redis_dsn%'
            logging: '%kernel.debug%'
            options:
                parameters:
                    database: 1
```

B-Galati:
Hi and thanks,

I will look into it, in the meantime you can try with DSN based solution which should work.

peter-gribanov:
> you can try with DSN based solution which should

Yes. It works. I can specify the database number in each DSN. But in this case, i lose the elasticity of the configuration.
The database number is the application level configuration.
The DSN for Redis is the configuration of the environment level.
I have 10 databases . If i configure the database number via environment variables, then i will have to configure 10 environment variables instead of one. This is inconvenient and high probability of errors.

B-Galati:
Thank you!

PR #556, opened 2020-03-02
Support for connection parameters

DemigodCode:
In newer versions connection parameters have to be in the options layer.
For bc the parameters layer not removed.

Related: https://github.com/snc/SncRedisBundle/issues/513


DemigodCode:
There's a compatibility problem(return value definition) with the new symfony version, don't know how to fix it without loading different classes for different kernel versions.

B-Galati:
Closing. 2.X version is not maintained anymore.

PR #567, opened 2020-03-20
Update 2.0.0 upgrade notes

gigo6000:
Mention that is required to use a `handler_id` if you are using Redis sessions.

We just upgraded a project from `1.1.9` to `2.0.0` and this broke our sessions.


B-Galati:
Thanks @gigo6000

PR #571, opened 2020-04-10
remove redundant variable

dzarda86:


curry684:
Given that the docs require the iterator to be the first variable - did this just never work before or what?

https://github.com/phpredis/phpredis#scan

dzarda86:
As depencency in our projects it fails to pass coding standards as the $key parameter is not defined anywhere in the function. We are currently unable to update to 3.* version.

snc/redis-bundle/Client/Phpredis/Client.php
/**
 * {@inheritdoc}
 */
public function scan(&$iterator, $pattern = null, $count = 0)
{
    return $this->call('scan', array($key, &$iterator, $pattern, $count));
}

B-Galati:
But the bugfix targets 2.1 branch :thinking: 

What's the current version of the bundle you are using?

dzarda86:
May be I did that wrong..

I am using composer package based on 2.1.13 tag; this issue is present since 2.1.10.

I was hoping you would merge this fix and release a new package version.

B-Galati:
We could but the CI is completely broken :thinking:

PR #573, opened 2020-04-22
fix: ClientCluster::flushall compatibility with redis extension

### LINKED ISSUE #572: Declaration ClientCluster::flushAll not compatible
The declaration of ClientCluster::flushall function is not compatible with the one defined in the extension (pecl redis 5.2.1): [redis_cluster.c](https://github.com/phpredis/phpredis/blob/6808cd6a7a30c006bc9133c3058dfe18551d7734/redis_cluster.c#L78)

> Warning: Declaration of ClientCluster::flushAll(...$args) should be compatible with RedisCluster::flushall($key_or_address, $async = NULL) in ..

Shouldn't it be ?

`  public function flushAll($key_or_address, $async = NULL)
    {
        return $this->call('flushAll', func_get_args());
    }`

firnas:
Fixes #572

### DIFF
--- a/Client/Phpredis/ClientCluster.php
+++ b/Client/Phpredis/ClientCluster.php
@@ -893,9 +893,9 @@ class ClientCluster extends RedisCluster
     /**
      * {@inheritdoc}
      */
-    public function flushAll(...$args)
+    public function flushAll($key_or_address, $async = NULL)
     {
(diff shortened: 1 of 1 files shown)


B-Galati:
Thanks @firnas

PR #577, opened 2020-05-17
Added magic method solved problem with zAdd

AndyDune:
Here is my error:

`
Declaration of Snc\RedisBundle\Client\Phpredis\Client::zAdd($key, $score1, $value1, $score2 = NULL, $value2 = NULL, $scoreN = NULL, $valueN = NULL) should be compatible with Redis::zAdd($key, $score, $value, ...$extra_args)
`
Why it does not use magic function? It resolves problem like this in future.
And described method in annotation.

### DIFF
--- a/Client/Phpredis/Client.php
+++ b/Client/Phpredis/Client.php
@@ -27,7 +27,7 @@ use Snc\RedisBundle\Logger\RedisLogger;
  * @author Yassine Khial <yassine.khial@blablacar.com>
  * @author Pierre Boudelle <pierre.boudelle@gmail.com>
  */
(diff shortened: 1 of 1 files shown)


B-Galati:
Sadly this approach does not work as it's then impossible to type the client with `\Redis` in userland code.

curry684:
As stated we have to extend `\Redis` for drop-in replacement purposes. So `zAdd` has a definition, and the magic method will not be invoked.

Best (and only) solution would be to generate the proxy on the fly but nobody has had time or interest to build that yet.

Closing as the proposed solution will not work.

PR #579, opened 2020-05-26
make phpredis work with lazy doctrine caches

### LINKED ISSUE #578: Lazy config is not working with phpredis.
I've noticed that a lazy config with phpredis and doctrine still requires a connection to the server, otherwise it fails:

I ran a `bin/console cache:clear` in my Dockerfile and noticed this bug.

The `snc_redis.<name>` services are lazy, but the `\Snc\RedisBundle\DependencyInjection\SncRedisExtension::loadDoctrine` provides this `$definitionFunction`:

```php
case 'phpredis':
    $definitionFunction = function ($client, $cache) use ($container) {
    $def = new Definition($container->getParameter('snc_redis.doctrine_cache_phpredis.class'));
    $def->addMethodCall('setRedis', array($client));
    if ($cache['namespace']) {
      $def->addMethodCall('setNamespace', array($cache['namespace']));
    }
    return $def;
  };
  break;
```
Seems okay and somehow "lazy", but once the actual `\Doctrine\Common\Cache\RedisCache::setRedis` call is made (e.g. during container warmup) it calls `$redis->setOption(Redis::OPT_SERIALIZER, $this->getSerializerValue());` on the proxy which will be instantiated immediately, which calls `\Snc\RedisBundle\Factory\PhpredisClientFactory::createClient` which includes this:

```php
if (!empty($options['connection_persistent'])) {
 $client->pconnect(...$connectParameters);
} else {
 $client->connect(...$connectParameters);
}
```
...but there is no connection. :(

I'll provide a pull request with a solution in a few minutes.

Hotfix in my app: Manually tagging all `doctrine.orm.**_cache` definitions lazy in a CompilerPass.

Basster:
Hopefully fixes #578

### DIFF
--- a/Client/Phpredis/Phpredis.php
+++ b/Client/Phpredis/Phpredis.php
@@ -1,0 +1,53 @@
+<?php
+declare(strict_types=1);
+
+namespace Snc\RedisBundle\Client\Phpredis;
+
+/**
+ * @internal
+ *
+ * Utility class for PHP Redis tasks
+ *
+ * @author Ole Rößner <oroessner@gmail.com>
+ */
+final class Phpredis
+{
+    /**
+     * Older version of phpredis extension do not support lazy loading
+     */
+    private const MINIMUM_LAZY_VERSION = '4.1.1';
+
+    /** @var string|null redis extension version */
+    private $version;
+
+    private $supportsLazy;
+
+    public function supportsLazyServices(): bool
+    {
+        if (null === $this->supportsLazy) {
+            $this->supportsLazy = $this->versionIsGreaterOrEqual(self::MINIMUM_LAZY_VERSION);
+            if (false === $this->supportsLazy) {
+                @trigger_error(
+                  sprintf('Lazy loading Redis is not supported on PhpRedis %s. Please update to PhpRedis %s or higher.',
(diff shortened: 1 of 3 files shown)


B-Galati:
Closing as per issue comment saying that it is fixed.

progmancod:
Guys, this is a NO solution. The solution in the comment issue take doctrine system cache out from redis at all:

doctrine.system_cache_pool:
adapter: cache.system

We still need this pull request.

B-Galati:
What's the issue with that?

progmancod:
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Doctrine\Common\Cache\RedisCache::setRedis() must be an instance of Redis, instance of Container1WKlMkT\RedisCluster_8dcfeee given, called in /var/www/html/var/cache/prod/Container1WKlMkT/SuppCore_AdministrativoBackend_KernelProdDebugContainer.php on line 1568 and defined in /var/www/html/vendor/doctrine/cache/lib/Doctrine/Common/Cache/RedisCache.php:31

B-Galati:
It looks like you're using RedisCluster while doctrine cache does not support it.

I don't get how it relates. Please open a detail issue if you think there is one.

PR #580, opened 2020-06-01
Deprecation since Symfony 5.1: NodeDefinition::setDeprecated() requires 3 arguments

StudioMaX:
The signature of method `NodeDefinition::setDeprecated()` has been updated to `NodeDefinition::setDeprecation(string $package, string $version, string $message)`.

See https://github.com/symfony/symfony/blob/5.1/UPGRADE-5.1.md

### DIFF
--- a/DependencyInjection/Configuration/Configuration.php
+++ b/DependencyInjection/Configuration/Configuration.php
@@ -11,6 +11,7 @@
 
 namespace Snc\RedisBundle\DependencyInjection\Configuration;
 
+use Symfony\Component\Config\Definition\BaseNode;
 use function method_exists;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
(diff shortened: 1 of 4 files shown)


DependencyInjection/Configuration/Configuration.php
```diff
@@ -267,7 +268,7 @@ private function addProfilerStorageSection(ArrayNodeDefinition $rootNode)
         $rootNode
             ->children()
                 ->arrayNode('profiler_storage')
-                    ->setDeprecated('Redis profiler storage is not available anymore since Symfony 4.4')
+                    ->setDeprecated(...$this->getProfilerStorageDeprecationMessage())
```
j0k3r, line 271: This isn't going to work on PHP < 7.4 right?

StudioMaX: Sorry, I don't get it, you mean [argument unpacking](https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list) via `...`? It will work since PHP 5.6 - https://3v4l.org/XSsm4

j0k3r: hu ho, sorry I wasn't aware of that.
I was thinking of https://wiki.php.net/rfc/spread_operator_for_array

B-Galati:
Thanks @StudioMaX

Issue #585, opened 2020-07-11
How can we configure RedisCluster options?


renatogcarvalho:
Hello folks,

I do have a Master/Slave replication on an environment in AWS with two different endpoints that I can seamlessly connect to a writer or a set of readers (which I can autoscale depending on load) using Predis with the replication option set to true. 

However, as per stated in your documentation and due to the inactivity in the Predis library, I decided to do some more reading about phpredis (specifically in regards to the RedisCluster implementation) and I looks like even though we don't have a "replication option" (i.e. https://github.com/phpredis/phpredis/issues/663) we might be able to accomplish a similar result by setting some options in the RedisCluster as per their specs: https://github.com/phpredis/phpredis/blob/develop/cluster.markdown#automatic-slave-failover--distribution

I'm specifically interested on this option:

```
// Always distribute readonly commands to the slaves, at random
$obj_cluster->setOption(
    RedisCluster::OPT_SLAVE_FAILOVER, RedisCluster::FAILOVER_DISTRIBUTE_SLAVES
);
```

How can we set these options via the SncRedis config? For example, for predis we're able to pass an array of options to define profile, connection_timeout, read_write_timeout, etc. Or perhaps the library should allow defining a custom "type" wrapper so we can set this configuration manually?

Thank you.

-R.

renatogcarvalho:
Maybe this can be accomplished by updating the PhpredisClientFactory's createClusterClient method with:

```
...

if (isset($options['slave_failover'])) {
    $client->setOption(\RedisCluster::OPT_SLAVE_FAILOVER, $this->loadSlaveFailoverType($options['slave_failover']));
}

return $client;
``` 

```
/**
 * Load the correct slave failover for RedisCluster
 *
 * @param string $type
 *
 * @return string
 * @throws InvalidConfigurationException
 */
private function loadSlaveFailoverType($type)
{
    $types = [
        'none' => \RedisCluster::FAILOVER_NONE,
        'error' => \RedisCluster::FAILOVER_ERROR,
        'distribute' => \RedisCluster::FAILOVER_DISTRIBUTE,
        'distribute_slaves' => \RedisCluster::FAILOVER_DISTRIBUTE_SLAVES
    ];

    if (array_key_exists($type, $types)) {
        return $types[$type];
    }

    throw new InvalidConfigurationException(sprintf('%s in not a valid slave failover. Valid failovers: %s', $type, implode(', ', array_keys($types))));
}
```

[...]

PR #586, opened 2020-07-11
Enable slave_failover option for phpredis \RedisCluster client

renatogcarvalho:
Q | A
-- | --
Branch | master
Bug fix? | no
New feature? | yes
BC breaks? | no
Deprecations? | no
Tests pass? | yes
Fixed tickets | n/a
License | MIT
Doc PR | n/a

Update for #585


Resources/config/schema/redis-1.0.xsd
```diff
@@ -72,6 +73,15 @@
         </xsd:restriction>
     </xsd:simpleType>
 
+    <xsd:simpleType name="phpredis-slave-failover-type">
```
B-Galati, line 76: I did not know about this file :open_mouth: 

@curry684 Should we continue to maintain it? Is this an old best practice for bundles?

B-Galati:
LGTM, could you also add some tests in `PhpredisClientFactoryTest` please? A test within the extension with this new config would be great too.

Resources/config/schema/redis-1.0.xsd
curry684: Can't say I've ever seen it either haha. Wasn't even aware that Symfony had XSD support for this, I was always under the assumption that the TreeBuilder had been around for that since ages past.

I think it's a relic from the really old days, before config went to YAML and now PHP itself. It's likely immensely outdated as well.

B-Galati:
@curry684 Thank you! :sparkles:

Resources/config/schema/redis-1.0.xsd
B-Galati: Thank you :-)

B-Galati:
@renatogcarvalho Could you add some tests please?

Issue #592, opened 2020-09-28
AUTH initial command missing when options:parameters:password is specified


zexa:
Hi, 

It seems like the following settings do not add an `AUTH` initial command

```
snc_redis:
  clients:
    default:
      dsn: "redis://127.0.0.1:8999"
      options:
        parameters:
          password: "passwordInParameters"
```

But these do.

```
snc_redis:
  clients:
    default:
      dsn: "redis://passwordInDSN@127.0.0.1:8999"
```

The expected behavior was that both of these settings styles would do the exact same thing. 

Is the style in the first example deprecated? If not what can I do to make it work?

B-Galati:
Hi, Both of them should work indeed. Don't know which one is the priority tough

PR #593, opened 2020-10-02
Specify the password via parameters for non sentinel connections

zexa:
Before this change, the options:parameters:password was reserved for sentinel
connections only, as we need a way to pass the auth to the servers that
sentinel provides for us, but not to sentinel itself.

This change allows us to use options:parameters:password for regular redis
connections.

The code checks if replication is set, because we want to keep the sentinel
functionality as is, which in this case is to leave the password in parameters
and allow sentinel to build the connection later.

### DIFF
--- a/Factory/PredisParametersFactory.php
+++ b/Factory/PredisParametersFactory.php
@@ -40,7 +40,8 @@ class PredisParametersFactory
         }
 
         if (
-            !isset($finalOptions['replication'])
+            !isset($finalOptions['password'])
+            && !isset($finalOptions['replication'])
             && isset($finalOptions['parameters']['password'])
         ) {
             $finalOptions['password'] = $finalOptions['parameters']['password'];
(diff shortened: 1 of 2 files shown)


zexa:
Related to https://github.com/snc/SncRedisBundle/issues/592

B-Galati:
Can you add some tests please?

B-Galati:
@zexa I've just seen that you targeted 2.1 branch but it is not maintained anymore. Can you target master please?

B-Galati:
It has been closed because I deleted 2.1 branch, please re-open against master ;-)

zexa:
I specificly targeted the 2.1 branch because the company I work at uses php 7.0 which version 3.x does not support (Minimum php version 7.1 AFAIK). 

If I apply a patch on master, will a tag be made for 2.1?

B-Galati:
no sorry, you must update

PR #594, opened 2020-10-20
Support nested DSNs generated by process environment variables

nesk:
I was trying to use this config for a project:

```yaml
snc_redis:
    clients:
        default:
            type: phpredis
            alias: default
            dsn: "%env(csv:REDIS_URL)%"
            options:
                cluster: true
```

This allows to use a simple list of redis instances in one env var:

```
REDIS_URL=redis://host_1,redis://host_2
```

However, when the processed environment variable returns an array, the `PhpredisClientFactory` receives the following DSN config:

```php
array (
  0 => 
  array (
    0 => 'redis://host_1',
    1 => 'redis://host_2',
  ),
)
```

This is difficult to solve in the configuration processor because the environment variable is stored as a simple string placeholder and not with the real and final value, so we can't detect the type of the value.

The solution is to detect, inside the `PhpredisClientFactory` class, if the DSN config contains only one item and if the first item is an array, then we simply use the nested array as the whole DSN config.

### DIFF
--- a/Factory/PhpredisClientFactory.php
+++ b/Factory/PhpredisClientFactory.php
@@ -44,6 +44,9 @@ class PhpredisClientFactory
             throw new \LogicException(sprintf('The factory can only instantiate \Redis|\RedisCluster classes: "%s" asked', $class));
         }
 
+        // Normalize the DSNs, because using processed environment variables could lead to nested values.
+        $dsns = count($dsns) === 1 && is_array($dsns[0]) ? $dsns[0] : $dsns;
+
         $parsedDsns = array_map(static function (string $dsn) {
             return new RedisDsn($dsn);
         }, $dsns);
--- a/Tests/Factory/PhpredisClientFactoryTest.php
+++ b/Tests/Factory/PhpredisClientFactoryTest.php
@@ -110,4 +110,26 @@ class PhpredisClientFactoryTest extends TestCase
         $this->assertSame('pass', $client->getAuth());
         $this->assertNull($client->getPersistentID());
     }
+
+    public function testNestedDsnConfig()
+    {
+        $factory = new PhpredisClientFactory();
+
+        $client = $factory->create(
+            \Redis::class,
+            [['redis://redis:pass@localhost:6379/2']],
+            array(
+                'parameters' => [
+                    'database' => 3,
+                    'password' => 'secret',
+                ],
+            ),
+            'alias_test'
+        );
+
+        $this->assertInstanceOf(\Redis::class, $client);
+        $this->assertSame(2, $client->getDBNum());
+        $this->assertSame('pass', $client->getAuth());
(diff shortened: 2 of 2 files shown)


nesk:
Any news? I'd rather prefer a hard no than no answer. 😉

B-Galati:
> Any news? I'd rather prefer a hard no than no answer. wink

Sorry I missed that one so thanks for the heads up.

Factory/PhpredisClientFactory.php
```diff
@@ -44,6 +44,9 @@ public function create($class, array $dsns, $options, $alias)
             throw new \LogicException(sprintf('The factory can only instantiate \Redis|\RedisCluster classes: "%s" asked', $class));
         }
 
+        // Normalize the DSNs, because using processed environment variables could lead to nested values.
+        $dsns = count($dsns) === 1 && is_array($dsns[0]) ? $dsns[0] : $dsns;
```
B-Galati, line 48: Is this the way it is managed in other bundle or that's more like a hack? 

Should we fix the config of the bundle instead?

B-Galati (approved):
LGTM thanks!

B-Galati:
Could you rebase the branch so that we see if it's the CI pass?

Factory/PhpredisClientFactory.php
nesk: This is more like a hack. Unfortunately I didn't find any other way to solve this…

When we write this in the config :

```
dsn: "%env(csv:REDIS_URL)%"
```

The `SncRedisExtension` class only sees an array with a single string, representing an env var placeholder. There's only the `PhpredisClientFactory` class which can see the final values.

nesk:
The branch has been rebased

Factory/PhpredisClientFactory.php
B-Galati: ok thank you!

B-Galati:
Thanks @nesk

Factory/PhpredisClientFactory.php
nesk: Oooh, maybe we could solve this directly in the config with [`ContainerBuilder::resolveEnvPlaceholders()`](https://github.com/symfony/symfony/blob/0dce5cf0ba439450475519137541e7f48894852f/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L1352)…

I will try to look for a solution with this method.

B-Galati: No because it would mean the cache is built with resolved environment instead of resolving them at runtime. We already bugs related to this behavior.

nesk: Nevermind then. Lets keep the hacky thing. 🙂

B-Galati:
release in 3.3.0 FYI! Sorry for the long delay :sweat_smile:

PR #599, opened 2020-11-29
Add support for PHP 8

nicwortel:
This includes upgrading to PHPUnit 9 for PHP 8 support. As PHPUnit 9 only supports PHP >= 7.3, support for older PHP versions is dropped as they can no longer be automatically tested against.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: php
 
-services:
-  - redis-server
-
 matrix:
   fast_finish: true
   include:
@@ -48,6 +45,7 @@ before_install:
   - composer global require --no-progress --no-scripts --no-plugins symfony/flex
   - if ! [ -z "${STABILITY}" ]; then composer config minimum-stability ${STABILITY}; fi;
   - if [ "${SYMFONY_VERSION}" != "" ]; then composer config extra.symfony.require ${SYMFONY_VERSION}; fi;
+  - sudo redis-server --requirepass sncredis --daemonize yes
   - for PORT in {7000..7008}; do sudo redis-server --port ${PORT} --cluster-enabled yes --cluster-config-file ${PORT}.conf --daemonize yes; echo 127.0.0.1:${PORT}; done
   - echo yes | redis-cli --cluster create $(seq -f 127.0.0.1:%g 7000 7008) --cluster-replicas 2
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,15 +12,25 @@ matrix:
       env: SYMFONY_VERSION="4.4.*"
     - php: 7.4
       env: SYMFONY_VERSION="4.4.*"
+    - php: 8.0
+      env: SYMFONY_VERSION="4.4.*"
     - php: 7.4
+      env: SYMFONY_VERSION="5.0.*"
+    - php: 8.0
       env: SYMFONY_VERSION="5.0.*"
     - php: 7.4
       env: SYMFONY_VERSION="5.1.*"
+    - php: 8.0
+      env: SYMFONY_VERSION="5.1.*"
     - php: 7.4
+      env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
+    - php: 8.0
       env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
 
   allow_failures:
     - php: 7.4
+      env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
+    - php: 8.0
       env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
 
 dist: bionic
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,18 +6,14 @@ services:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
+    - php: 7.3
       env: SYMFONY_VERSION="3.4.*" COMPOSER_FLAGS="--prefer-lowest"
-    - php: 7.2
-      env: SYMFONY_VERSION="4.4.*"
     - php: 7.3
       env: SYMFONY_VERSION="4.4.*"
     - php: 7.4
       env: SYMFONY_VERSION="4.4.*"
     - php: 7.4
       env: SYMFONY_VERSION="5.0.*"
-    - php: 7.2
-      env: SYMFONY_VERSION="5.1.*"
     - php: 7.4
       env: SYMFONY_VERSION="5.1.*"
     - php: 7.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,10 @@ language: php
 matrix:
   fast_finish: true
   include:
-    - php: 7.3
+    - php: 7.2
       env: SYMFONY_VERSION="3.4.*" COMPOSER_FLAGS="--prefer-lowest"
+    - php: 7.2
+      env: SYMFONY_VERSION="4.4.*"
     - php: 7.3
       env: SYMFONY_VERSION="4.4.*"
     - php: 7.4
@@ -15,6 +17,8 @@ matrix:
       env: SYMFONY_VERSION="5.0.*"
     - php: 8.0
       env: SYMFONY_VERSION="5.0.*"
+    - php: 7.2
+      env: SYMFONY_VERSION="5.1.*"
     - php: 7.4
       env: SYMFONY_VERSION="5.1.*"
     - php: 8.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,15 +23,9 @@ matrix:
     - php: 8.0
       env: SYMFONY_VERSION="5.1.*"
     - php: 7.4
-      env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
+      env: SYMFONY_VERSION="5.2.*"
     - php: 8.0
-      env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
-
-  allow_failures:
-    - php: 7.4
-      env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
-    - php: 8.0
-      env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
+      env: SYMFONY_VERSION="5.2.*"
 
 dist: bionic
 sudo: false
--- a/DependencyInjection/Compiler/SwiftMailerPass.php
+++ b/DependencyInjection/Compiler/SwiftMailerPass.php
@@ -36,7 +36,7 @@ class SwiftMailerPass implements CompilerPassInterface
                 // default class, lets check for Swift Mailer version and set correct class
                 $class = new \ReflectionClass('\\Swift_Spool');
                 $parameters = $class->getMethod('queueMessage')->getParameters();
-                switch ($parameters[0]->getClass()->getName()) {
+                switch ($parameters[0]->getType()) {
                     case 'Swift_Mime_Message':
                         // Swift Mailer 5.x
                         $handlerDefinition->setClass($handlerDefinition->getClass().'5');
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^7.2 || ^8.0",
         "symfony/framework-bundle": "^3.4.23 || ^4.2.4 || ^5.0",
         "symfony/http-foundation": "^3.4 || ^4.2 || ^5.0",
         "symfony/yaml": "^3.4 || ^4.2 || ^5.0"
@@ -29,7 +29,7 @@
         "ext-pdo_sqlite": "*",
         "doctrine/doctrine-bundle": "^1.11.2 || ^2.0",
         "doctrine/orm": "^2.6.4",
-        "phpunit/phpunit": "^9.4",
+        "phpunit/phpunit": "^8.5",
         "predis/predis": "^1.1",
         "swiftmailer/swiftmailer": "^6.0",
         "symfony/browser-kit": "^3.4 || ^4.2 || ^5.0",
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,8 @@
         }
     ],
     "require": {
-        "php": "^7.1.3",
-        "symfony/framework-bundle": "^3.4 || ^4.2 || ^5.0",
+        "php": "^7.3 || ^8.0",
+        "symfony/framework-bundle": "^3.4.23 || ^4.2.4 || ^5.0",
         "symfony/http-foundation": "^3.4 || ^4.2 || ^5.0",
         "symfony/yaml": "^3.4 || ^4.2 || ^5.0"
     },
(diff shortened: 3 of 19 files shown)


nicwortel:
Currently the PHP 8.0 jobs are still failing because of:

- [Travis CI not supporting PHP 8.0 builds](https://travis-ci.community/t/php-8-0-missing/10132/10?u=nicwortel)
- [doctrine/orm doesn't have a version released yet which supports PHP 8.0](https://github.com/doctrine/orm/issues/8350)

curry684:
> doctrine/orm doesn't have a version released yet which supports PHP 8.0

That alone is a pretty good reason to park this for a bit still   😉

nicwortel:
doctrine/orm [2.8.0](https://github.com/doctrine/orm/releases/tag/2.8.0) has now been released which supports PHP 8.0.

B-Galati:
Does Travis supports PHP 8.0 now?

.travis.yml
```diff
@@ -6,26 +6,32 @@ services:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
+    - php: 7.3
```
B-Galati, line 9: Please keep PHP 7.1. We are still supporting this version of PHP

.travis.yml
```diff
@@ -6,26 +6,32 @@ services:
...
-    - php: 7.2
-      env: SYMFONY_VERSION="5.1.*"
+    - php: 8.0
+      env: SYMFONY_VERSION="5.0.*"
     - php: 7.4
       env: SYMFONY_VERSION="5.1.*"
+    - php: 8.0
+      env: SYMFONY_VERSION="5.1.*"
     - php: 7.4
       env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
```
B-Galati, line 26:
```suggestion
      env: SYMFONY_VERSION="5.2.*"
```

.travis.yml
```diff
@@ -6,26 +6,32 @@ services:
...
+    - php: 8.0
+      env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
```
B-Galati, line 28: Symfony 5.2 is now stable.

```suggestion
      env: SYMFONY_VERSION="5.2.*"
```

.travis.yml
```diff
@@ -6,26 +6,32 @@ services:
...
 
   allow_failures:
     - php: 7.4
       env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
+    - php: 8.0
+      env: SYMFONY_VERSION="5.2.*" STABILITY="dev"
```
B-Galati, line 34: If we want to check the next version of Symfony, matrix update also needed.

```suggestion
      env: SYMFONY_VERSION="5.3.*" STABILITY="dev"
```

composer.json
```diff
@@ -20,23 +20,23 @@
         }
     ],
     "require": {
-        "php": "^7.1.3",
-        "symfony/framework-bundle": "^3.4 || ^4.2 || ^5.0",
+        "php": "^7.3 || ^8.0",
```
B-Galati, line 23:
```suggestion
        "php": "^7.1.3 || ^8.0",
```

composer.json
```diff
@@ -20,23 +20,23 @@
...
+        "symfony/framework-bundle": "^3.4.23 || ^4.2.4 || ^5.0",
```
B-Galati, line 24: Can be reverted IMHO.

nicwortel:
> Does Travis supports PHP 8.0 now?

According to https://travis-ci.community/t/php-8-0-missing/10132, yes. We'll see when I push a new commit.

.travis.yml
nicwortel: Is it a deliberate decision to continue support of outdated PHP versions? PHP 7.1 and 7.2 are [end-of-life](https://www.php.net/eol.php). They are no longer receiving security updates and installations should be upgraded to newer PHP versions as soon as possible.

On top of that, PHP 7.1 and 7.2 are [not supported by the latest PHPUnit](https://phpunit.de/supported-versions.html), PHPUnit 9. I could use a wider version constraint for PHPUnit (something like `^7.5 || ^8.0 || ^9.0`) to allow tests on these PHP versions, but it might add some additional complexity. Of course, if support for these PHP versions really is required this is the only way to go.

.travis.yml
nicwortel: I think this is out of scope for this PR, but I will create an issue to track this.

composer.json
nicwortel: Older versions of Symfony had a KernelTestCase which was incompatible with PHPUnit 8. This broke the Travis job which runs with `--prefer-lowest`. See https://travis-ci.org/github/snc/SncRedisBundle/jobs/746572032, https://github.com/snc/SncRedisBundle/pull/599/commits/ae42ace2caee86f0ad251852ac835e8d37de2d68 and https://github.com/symfony/symfony/commit/1077df60ab35c418f8ca196db40500f5489990cd.

composer.json
nicwortel: See https://github.com/snc/SncRedisBundle/pull/599#discussion_r538197563.

.travis.yml
B-Galati: Interesting thanks! @curry684 what's your opinion?

.travis.yml
B-Galati: thanks

composer.json
B-Galati: Makes sense, thanks.

.travis.yml
B-Galati: I know the Symfony PhpUnit Bridge could solve that issue but I am not a fan of the way it works.

nicwortel:
PHP 8.0 is working on Travis CI now but the PHPUnit tests fail with:
> RedisException: ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?

I'm not sure what's causing this or why it stopped working with PHP 8.0. I would have to dig deeper into this but I currently don't have a lot of time to do so.

B-Galati:
> PHP 8.0 is working on Travis CI now but on PHP 8.0 [the PHPUnit tests fail with](https://travis-ci.org/github/snc/SncRedisBundle/jobs/748309926):
> 
> > RedisException: ERR AUTH  called without any password configured for the default user. Are you sure your configuration is correct?
> 
> I'm not sure what's causing this or why it stopped working with PHP 8.0. I would have to dig deeper into this but I currently don't have a lot of time to do so.

Yeap that's not an easy problem, that's a behavior change related to the last PhpRedis version IIRC. 

Don't know what to do about it yet.

nicwortel:
> Yeap that's not an easy problem, that's a behavior change related to the last PhpRedis version IIRC.

Do you have a reference to release notes or the version number of when this change was introduced? Then I'll take a look when I find some time.

B-Galati:
> > Yeap that's not an easy problem, that's a behavior change related to the last PhpRedis version IIRC.
> 
> Do you have a reference to release notes or the version number of when this change was introduced? Then I'll take a look when I find some time.

Nothing sorry 😐

nicwortel:
I took some time to debug the failing build.

Some of the tests (the `PhpredisClientFactoryTest` and the `SncRedisExtensionTest`) test setting a password in the bundle configuration, even though the Redis instance running in Travis CI does not have a password configured.

From what I can see testing this locally with Redis Docker images, it seems like Redis 5 didn't really care if you tried to authenticate with a password when no password was configured, but Redis 6 returns the "ERR AUTH called without any password configured for the default user. Are you sure your configuration is correct?" message.
I was able to trace this back to [this commit in Redis](https://github.com/redis/redis/commit/b87815c1f800690c90bbc6c50bbe62878841d0b9) which was indeed released with Redis 6.

I haven't been able to find the change in PhpRedis that would explain why the build succeeds for earlier PHP versions, but I think it's safe to say that tests with a password shouldn't run against a passwordless Redis installation.

B-Galati:
@nicwortel Thanks for the very clear explanation!
> I haven't been able to find the change in PhpRedis that would explain why the build succeeds for earlier PHP versions, but I think it's safe to say that tests with a password shouldn't run against a passwordless Redis installation.

I agree :+1: 

> I think we should either:
> 
>     * set a password and update all test code to authenticate against Redis
It's the best solution IMHO.
>     * not set a password, and skip the tests that try to authenticate against Redis

> 
> P.S. Happy holidays! snowman_with_snow

Thanks! You too!

B-Galati:
Is it ready @nicwortel? Could you clean-up your commits if yes?

nicwortel:
I will do it tonight hopefully, if I find the time :+1:

nicwortel:
> I will do it tonight hopefully, if I find the time :+1: 

Well we all know what happened last night :clown_face:, which got me a little distracted.

Anyway, I rebased my commits, squashing the ones that could be merged together. I'm awaiting the build result (the tests pass locally) but as far as I'm concerned, this PR can be reviewed.

Talking about CI builds, I noticed that this repository runs on travis-ci.org which will be shut down within several weeks from now:

afbeelding

While all repositories should automatically be migrated to travis-ci.com, Travis CI has changed its pricing model and no longer supports open source repositories for free (you can jump through a lot of hoops to apply for OSS credits but personally I'm still waiting for a response to the email I sent them in November last year...).

For this reason a lot of OSS repositories have migrated their CI pipeline to GitHub Actions and I recommend doing the same. Would you like me to open an issue for this?

.travis.yml
Seldaek: Now that PHPUnit 8.5 allow PHP 8, I believe it should be possible to run on phpunit `^7.5 || ^8.5` from php 7.1 to 8, with the same tests. Technically as long as the bundle supports symfony 4 IMO php 7.1 should be supported too.

DependencyInjection/Compiler/SwiftMailerPass.php
```diff
@@ -36,7 +36,7 @@ public function process(ContainerBuilder $container)
                 // default class, lets check for Swift Mailer version and set correct class
                 $class = new \ReflectionClass('\\Swift_Spool');
                 $parameters = $class->getMethod('queueMessage')->getParameters();
-                switch ($parameters[0]->getClass()->getName()) {
+                switch ($parameters[0]->getType()) {
```
Seldaek, line 39: I'm not sure why this change was done, because it seems like it uses __toString of ReflectionType which is deprecated https://www.php.net/manual/en/reflectiontype.tostring.php

nicwortel: Interesting... I actually changed it because PHPUnit on PHP 8 reported `ReflectionParameter::getClass()` as deprecated. Unfortunately the failing build is no longer visible after I rebased my commits.
[The stubs in PhpStorm say the same](https://github.com/JetBrains/phpstorm-stubs/blob/ed3c896f6d7fef78dfd42366c8c3fd596676d75a/Reflection/ReflectionParameter.php#L126), but the PHP documentation [doesn't say anything about it](https://www.php.net/manual/en/reflectionparameter.getclass.php).

The PhpStorm stub points to ReflectionParameter::getType() as a replacement:

>     #[Deprecated(reason: "Use ReflectionParameter::getType() and the ReflectionType APIs should be usedinstead.", since: "8.0")]

However I missed the deprecation warning for the `__toString()` method. I'll have to look into this.

Seldaek: So it looks like one is deprecated in 8, and one in 7.4, but the one deprecated in 7.4 isn't in 8 anymore?! This is kinda bonkers.

https://3v4l.org/23vV9

Seldaek: In any case, it seems to me like this should maybe not belong in this PR to avoid blocking the php 8 support which it would appear works regardless of this change.

nicwortel: It seems to be caused by (preparation for) the introduction of union types in PHP 8. `ReflectionType` has become abstract with two classes extending it, [ReflectionNamedType](https://www.php.net/manual/en/class.reflectionnamedtype.php) (since 7.1) and [ReflectionUnionType](https://www.php.net/manual/en/class.reflectionuniontype.php) (since 8.0). Calling `ReflectionParameter::getType()` will return `ReflectionNamedType` if the type hint contains one type, `ReflectionUnionType` if it is a union type, or `null` if there is no type hint. `ReflectionUnionType` has a `getTypes()` method which returns a list of `ReflectionNamedType` instances. `ReflectionNamedType` has a `getName()` method which we are supposed to call instead of relying on `__toString()`.

Unfortunately the documentation on this appears to be incomplete - I will see if I am able to contribute a pull request to php/doc.

As we check the parameter type hint of a known class here (`Swift_Spool`) and expect it to use either `Swift_Mime_Message` or `Swift_Mime_SimpleMessage` as a type hint, I believe it's safe to expect an instance of `ReflectionNamedType`, so I could call `->getName()` on it. That way we don't have to rely on `__toString()`. However, a static analysis tool such as PHPStan would probably fall over this because strictly speaking, `getType()` is not guaranteed to return a `ReflectionNamedType` instance.

If we really want to be on the safe side I could first check if `getType()` returns an instance of `ReflectionNamedType` but as stated above, that should not occur under normal circumstances.

.travis.yml
curry684: In general for libraries like this I'm in favor of following PHP support life cycles, so drop a PHP version once it's clearly end of life. So not the next day, but by all means with the next big update like the one we're discussing here.

However, it should most of all match the project requirements in my opinion, so *if we declare to support it we should test for it*. Hence the Travis PHP versions should match the minimum PHP version in `composer.json`. If you remove 7.1 from the tests, change `composer.json` to 7.2+.

B-Galati: thanks for your insights @Seldaek and @curry684 ! Much appreciated.

I am in favor of dropping PHP 7.1 support and keep PHP 7.2. And thus it should be tested against this version in the CI.

nicwortel: > Hence the Travis PHP versions should match the minimum PHP version in `composer.json`. If you remove 7.1 from the tests, change `composer.json` to 7.2+.

I agree, I had already changed the PHP version requirement in `composer.json` to `"^7.3 || ^8.0"`. If the conclusion of this thread is to continue supporting PHP 7.2 and/or 7.1, I'll make sure that they are re-introduced in `composer.json` as well as in the CI configuration.

As per @Seldaek's comment, [the addition of PHP 8 support to PHPUnit 8.5](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-8.5.md#8512---2020-11-30) has simplified this a bit, leaving us with three options:

1. Keep this PR as it is (support PHP 7.3, 7.4 and 8.0 using PHPUnit 9)
2. Support PHP 7.2, 7.3, 7.4 and 8.0 using PHPUnit 8.5
3. Support PHP 7.1, 7.2, 7.3, 7.4 and 8.0 using PHPUnit 7.5|8.5

In case of option 2 or 3 I might have to revert some of the changes to the test code but that shouldn't be a big issue.
I can imagine how option 3 might introduce some deprecation catch-22's (where a method has been deprecated in 8.5 but no alternative exists yet in 7.5) but I haven't checked that yet so it may not be a problem at all.

Perhaps it's good to keep in mind that this decision will come back with each new release of PHP. When PHP 8.1 will be released at the end of this year, it will not be supported by PHPUnit 7 or 8, but only by PHPUnit 9 (or even 10, I'm not sure). At some point you will be forced to choose between supporting the latest PHP version and older, EOL versions.

B-Galati: `2.` looks like a good solution to me. Thanks @nicwortel :pray:

B-Galati:
> For this reason a lot of OSS repositories have migrated their CI pipeline to GitHub Actions and I recommend doing the same. Would you like me to open an issue for this?

@nicwortel Yes you can ;-)

B-Galati (approved):

B-Galati:
@Seldaek @curry684 Does it look OK to both of you?

Seldaek (approved):

curry684 (approved):

B-Galati:
Thanks @nicwortel!

B-Galati:
I'll try to do the release ASAP.

And BTW thanks @Seldaek and @curry684 for your reviews.

PR #614, opened 2021-07-08
Doctrine cache 2

### LINKED ISSUE #611: Doctrine Cache 2.x
Heya,
I just ran into an issue when trying to upgrade some dependencies:
SncRedis is not compatible with version 2 of Doctrine Cache (Doctrine\Common\Cache\RedisCache is missing).
So it would be good to restrict that in the composer file for the time being (or make it compatible which would be more work I think :) )

KDederichs:
Fixes #611 

I ported those 2 Classes over from the 1.x versions of the cache bundles.
As it's a straight copy with some slight modifications I'm not 100% sure how licensing works here, maybe someone can enlighten me on that :)

### DIFF
--- a/Cache/PredisCache.php
+++ b/Cache/PredisCache.php
@@ -1,0 +1,108 @@
+<?php
+
+
+namespace Snc\RedisBundle\Cache;
+
+
+use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\CacheProvider;
+use Predis\Client;
+
+class PredisCache extends CacheProvider
+{
+
+    /**
+     * @var Client
+     */
+    private $client;
+
+    /**
+     * @param Client $client
+     *
+     * @return void
+     */
+    public function __construct(Client $client)
+    {
+        $this->client = $client;
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


KDederichs:
Ok scratch that, aparently they deprecated it for PSR-6/16 which is implemented by the symfony cache, so lets use that instead.

KDederichs:
Any reason CI is not starting on this btw?

B-Galati:
Hello, I don't know, I am barely maintaining this project now :sweat_smile: 

Must related to the new policy of Travis.

Otherwise, I wanted to remove this feature from the bundle and just document how to configure all of this explicitly.

KDederichs:
I mean this PR basically removes it for doctrine >= 2.9 tbh.
Older versions would still work so you still have BC.

ostrolucky:
@B-Galati @snc @curry684 Can I be a co-collaborator in this project?

curry684:
Yes that would be awesome, both me and @B-Galati stepped in to fix a lot of outdated stuff for the 3.x transition, but I'm also very short on time past months. I'll drop @snc a line.

Issue #466, opened 2018-10-19
falsy "failed (MOVED)" error logs when using cluster mode (following #416)


ebarault:
Hi @curry684, @linasm83,

Recalling the discussion we had in https://github.com/snc/SncRedisBundle/pull/416

Regarding this comment: https://github.com/snc/SncRedisBundle/pull/416#issuecomment-384413772

We keep having the `......... failed (MOVED) ..........` errors in our logs for every single first redis commands in a sequence although the commands themselves work.

Set for example i do an EXISTS, then a SETEX, i'll have the error log only for the EXISTS command, because as with the redis-cmi with `-c` option, the client then follows the redirection directly with no warning.

By any chance could we work on something to get rid of these cluster-mode related falsy error logs? or turn them to an INFO log level message indicating the `MOVED` part, but without mentioning the `failed` part ?

curry684:
I'm open to all suggestions as I agree the erroneous logs should be removed.

ebarault:
i'll start digging the origin of these logs, do you have a rough idea where to begin?

linasm83:
Sorry for late response, kinda busy recent weeks, but will try to find some time to investigate more. Personally not getting these errors in production, but might be due log configuration.

B-Galati:
@curry684 @ebarault @linasm83 

Do you have any example of working cluster mode? Because the doc looks completely wrong by specifying -> `cluster: Snc\RedisBundle\Client\Predis\Connection\PredisCluster` for example.

```
cluster:
            type: predis
            alias: cluster
            dsn:
                - redis://127.0.0.1/1
                - redis://127.0.0.2/2
                - redis://pw@/var/run/redis/redis-1.sock/10
                - redis://pw@127.0.0.1:63790/10
            options:
                prefix: foo
                profile: 2.4
                connection_timeout: 10
                connection_persistent: true
                read_write_timeout: 30
                iterable_multibulk: false
                throw_errors: true
                cluster: Snc\RedisBundle\Client\Predis\Connection\PredisCluster
```

curry684:
Nope, never used clustering myself at all.

Lutacon:
@B-Galati do you have it working in clustering mode?

B-Galati:
@Lutacon no, I never tried with predis

Lutacon:
We have it running with AWS ElastiCache in clustered mode(4 nodes) with this config:
```
snc_redis:
  clients:
    redis:
      type: predis
      alias: redis
[...]

PR #616, opened 2021-08-06
Do not falsily classify MOVED responses in logs as errors

ostrolucky:
Fixes https://github.com/snc/SncRedisBundle/issues/466


B-Galati:
Thanks @ostrolucky

Issue #615, opened 2021-08-06
Logging in cluster mode works only for first query


ostrolucky:
In cluster mode, eg. with config such as
```yml
snc_redis:
    clients:
        default:
            type: predis
            alias: default
            dsn: [redis://shard01.foo.bar:6380, redis://shard01.foo.bar:6381]
            logging: true
            options:
                prefix: 'BL::'
                cluster: redis
```

Logger is called only once, with message such as 
```
HGET BL::shopping_wpset_pid_mapping default
An error occured: MOVED 1151 10.111.167.132:6381
```

After this, apparently what happens is that Predis creates new connection ([StreamConnection](https://github.com/predis/predis/blob/54775a02954ac425daef0c3c717665e968b5fc84/src/Connection/Factory.php#L26)) which doesn't use ConnectionWrapper.

This is because RedisCluster::createConnection at https://github.com/predis/predis/blob/54775a02954ac425daef0c3c717665e968b5fc84/src/Connection/Cluster/RedisCluster.php#L308 overwrites parameters which contain `logging: true` flag. This is checked at https://github.com/snc/SncRedisBundle/blob/d7718cd75166bb256380cf428cbde54402be149e/Client/Predis/Connection/ConnectionFactory.php#L64 but by then configured parameters are gone and there is only host and port left.

Additionally, there is wrong phpdoc with StreamWrapper assumption at https://github.com/snc/SncRedisBundle/blob/d7718cd75166bb256380cf428cbde54402be149e/Client/Predis/Connection/ConnectionFactory.php#L62, as it turned out after constructing it, it returns [StreamConnection and not ConnectionWrapper](https://github.com/predis/predis/blob/54775a02954ac425daef0c3c717665e968b5fc84/src/Connection/Factory.php#L26)

ostrolucky:
Workaround:
```php
<?php
declare(strict_types=1);

namespace App;

use Snc\RedisBundle\Logger\RedisLogger;

class ConnectionFactory extends \Snc\RedisBundle\Client\Predis\Connection\ConnectionFactory
{
    public function create($parameters)
    {
        $connection = parent::create($parameters);

        if (null == $this->wrapper) {
            return $connection;
        }

        $wrapper = $this->wrapper;
        $connection = new $wrapper($connection);
        $connection->setLogger($this->logger);

        return $connection;
    }
}
```

```yml
snc_redis:
    class:
        connection_factory: App\ConnectionFactory
    clients:
        default:
            type: predis
[...]

PR #619, opened 2021-08-07
Fix predis RedisCluster logging

ostrolucky:
When Predis initializes new connection from within RedisCluster class, it throws away all previous parameters. Hence we cannot rely them anymore.

This patch assumes that if logger and wrapper class have been injected, logging is desired.

Fixes https://github.com/snc/SncRedisBundle/issues/615


ostrolucky:
rebased, but your test suite fails for unrelated reasons

PR #622, opened 2021-08-12
RedisProfilerStorage: fix method signatures for Symfony 4.4

camilledejoye:
Hello,

When migrating to Symfony 4.4 I had errors regarding the signature of some methods in the `RedisProfilerStorage` class.
This PR simply fixes them, I didn't try to look for other inconsistencies but for now it seems to work.

### DIFF
--- a/Profiler/Storage/RedisProfilerStorage.php
+++ b/Profiler/Storage/RedisProfilerStorage.php
@@ -74,7 +74,7 @@ class RedisProfilerStorage implements ProfilerStorageInterface
     /**
      * {@inheritdoc}
      */
-    public function find($ip, $url, $limit, $method, $start = null, $end = null)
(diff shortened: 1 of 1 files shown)


Maxooo:
We have the same issue here, is it possible to take a look at this PR to merge it ?
Thank you @camilledejoye and thank you to @snc for your job on the bundle.

B-Galati:
Hello! I am sorry this feature is not supported anymore, see: 

https://github.com/snc/SncRedisBundle/blob/master/CHANGELOG.md#320-2019-11-14

>[BC BREAK] Deprecate support for redis profiler storage starting from Sf 4.4 (#543) - @B-Galati (7f4249)
>
>As per Symfony:
>
>    This interface exists for historical reasons. The only supported implementation is FileProfilerStorage. As the profiler must only be used on non-production servers, the file storage is more than enough and no other implementations will ever be supported.

PR #633, opened 2021-11-13
Run tests on PHP 8.1

ruudk:


ostrolucky:
Let's try to remove offending methods and see how many of them there are that are offending. Can you do that?

ruudk:
I would like to fix it but I'm not sure how. 

This PR installs the Redis extension from `develop` and they seem to have introduced strict types like https://github.com/phpredis/phpredis/blob/aac42cd33510030bb62973f8886f4cd98b9b8003/redis.stub.php and https://github.com/phpredis/phpredis/blob/aac42cd33510030bb62973f8886f4cd98b9b8003/redis_cluster.stub.php

Not sure what happens when you run it with the older untyped extension installed. Any advice? 😊

ostrolucky:
This needs to be finished to fix issues like this https://github.com/snc/SncRedisBundle/pull/511. I think until we have that, we won't have release for PHP 8.1

ostrolucky:
Superseded by https://github.com/snc/SncRedisBundle/pull/636

Issue #651, opened 2022-01-31
Version 4.1 proxy does not handle well sScan calls


magnetik:
With latest version 4.1.5. (It worked with 4.0)

All our calls to sScan fails when using the logging (because of the proxy).
Can this be because the first parameter should be passed by reference?

With 4.0, two consecutive calls of SSCAN will do :
```
 ['Executing command "SSCAN fleet 0 <null> <null>"'],
 ['Executing command "SSCAN fleet 0 <null> <null>"'],
```
First call return empty array
Second call return false

With 4.1 : 
```
 ['Executing command "SSCAN fleet <null> <null> <null>"'],
 ['Executing command "SSCAN fleet <null> <null> <null>"'],
```
Both calls return empty array.

it causes infinite loop using 4.1

Thanks

ostrolucky:
Hmm this is because of array_values call, which removes references. If someone has an idea how to reindex array while keeping references intact, please let us know.

ostrolucky:
See https://github.com/snc/SncRedisBundle/pull/653

magnetik:
Awesome thanks for your work!

PR #652, opened 2022-02-01
Test regarding bug with sccan

magnetik:
ref #651


ostrolucky:
Continues in https://github.com/snc/SncRedisBundle/pull/653

PR #659, opened 2022-04-24
Register Alias with an Argument for autowiring.

schodemeiss:
Allows auto-wiring of clients configured using this bundle via a named argument and the underlying type for that specific client.

See: https://symfony.com/doc/current/service_container.html#binding-arguments-by-name-or-type.

With a bundle configuration of:

``` yaml
snc_redis:
    clients:
        default:
            type: predis
            alias: default
            dsn: "%env(string:REDIS_URL)%"
```

You can autowire via:

```  php
public function __construct(Predis\Client $default) {
    ...
}
```


ostrolucky:
tests required

schodemeiss:
> tests required

Hmmm, not even sure where to begin testing that change. I guess I could write a test that checks to see if that definition is set in the container? Would that suffice?

ostrolucky:
Closing because of staleness

PR #662, opened 2022-05-09
Remove non UTF-8 chars from Stopwatch section names

### LINKED ISSUE #661: [Profiler] Don't log non UTF-8 output.
As I already mentioned in https://github.com/symfony/symfony/issues/46250 , sometimes the output logged to the Performance tab in the Profiler throws an error when the thing Snc Redis is trying to serialize is something containing non valid UTF-8.

Since I narrowed it down to what's causing it I thought I'd submit an issue/fix here for the time being till the folks over at symfony decide if that's something that can be tackled at the profiler level or is something for the bundles to keep in mind.

KDederichs:
Fixes #661

### DIFF
--- a/src/Logger/RedisCallInterceptor.php
+++ b/src/Logger/RedisCallInterceptor.php
@@ -39,7 +39,7 @@ class RedisCallInterceptor
         $time    = microtime(true);
 
         if ($this->stopwatch) {
-            $event = $this->stopwatch->start($this->makePrintable($command), 'redis');
+            $event = $this->stopwatch->start(preg_replace('/[^[:print:]]/', '', $command), 'redis');
         }
 
         $return = $instance->$method(...$args);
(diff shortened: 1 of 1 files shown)


KDederichs:
Hmm yeah fair, I think I found where to do it in SF so I'll close this and open a PR there!

Pazns:
@ostrolucky Nicolas Grekas suggested [here](https://github.com/symfony/symfony/pull/46303#issuecomment-1122104591) using this instead : `preg_match('//u', $string)`

KDederichs:
@ostrolucky 
Since it should be fixed here 2 options:
1) I use the `preg_match('//u', $string)` to detect it in flatten and do the same thing as already this PR already does but without dependency
2) I replace non printable characters in the stopwatch command only using `preg_replace` which one do you prefer?

ostrolucky:
2. pls

KDederichs:
There we go, let me know if something needs changing :)

src/Logger/RedisCallInterceptor.php
```diff
@@ -87,4 +88,19 @@ private function flatten(array $arguments, array &$list): void
...
+
+    /**
+     * Removes all non printable characters from input, returns default string if the whole input was non printable
+     *
+     * @param string $command The command to check
+     */
+    public function makePrintable(string $command): string
+    {
+        $sanitized = preg_replace('/[^[:print:]]/', '', $command);
+        if ($sanitized !== '') {
```
ostrolucky, line 100: This cannot happen, since part of $command is always method name, which is ASCII. Given this, I would suggest to inline preg_replace call directly.

ostrolucky (approved):

Issue #658, opened 2022-03-24
Add options to force TLS Version in the connection string


lbassin:
Hi guys,
I've been dealing with an issue that is not directly related to this bundle
If you want more details https://github.com/phpredis/phpredis/issues/1726 and https://bugs.php.net/bug.php?id=79501

TL;DR with PHP >= 7.4, tls v1.3 is used by default and seems to result in random frozen connections (Both Predis and PHPRedis sa it's at the php layer)
The quick fix in the meanwhile is to force the connection to use tls 1.2, this is achieved by using the following connection string `tlsv1.2://my-redis-server:6379` instead of `tls://my-redis-server:6379`

After looking at the factories, the change is pretty basic to do 
Here: https://github.com/snc/SncRedisBundle/blob/master/src/Factory/PhpredisClientFactory.php#L134
And here: https://github.com/snc/SncRedisBundle/blob/master/src/Factory/PredisParametersFactory.php#L49

But what I am not sure about and why I am creating this issue is to ask your opinion on how to provide this information in the snc bundle config.  
  
So far I've come up with 2 ideas:  
 
1. Add an option in the config (Easy and quick to do)
```yaml
snc_redis:
    clients:
        default:
            type: phpredis
            dsn: 'rediss://S3cR3t@127.0.0.1:6379'
            options:
                tls_version: 1.2
```
2. Add the information in the DSN (Requires extra parsing and "hard" to understand but closer to what is done by the bundle)
```yaml
snc_redis:
    clients:
        default:
            type: phpredis
            dsn: 'redissv1.2://S3cR3t@127.0.0.1:6379'
```

I'm gonna provide a MR for that, I just wanted your opinion first  
On my side I think I'm leaning more towards the 2nd option as it's closer to the reality even if more complex

ostrolucky:
Depends how likely is that this version differs between environments, eg. prod/dev. If it's unlikely, non-dns option is fine. Otherwise both should be supported, like with other connection related options

lbassin:
I think it's unlikely to have a different tls version between environments, so i'll start with the non-dsn option and i'll see for the other one later

ostrolucky:
But I think it's likely that prod will have TLS enabled, but dev not. What would happen in that case in dev if tls_version is specified?

lbassin:
I guess you are right this is something that can happen even if I don't really like it but anyway I had in mind to ignore the tls_version options if TLS is not enabled

ostrolucky:
[...]

PR #669, opened 2022-06-28
Add context configuration to phpredis

mattiabasone:
Add `context` configuration option for `connect/pconnect` calls.


succoDiPompelmo:
@ostrolucky any news on this PR ?

ostrolucky:
We had some discussion about this here https://github.com/snc/SncRedisBundle/issues/658 and I would prefer if this kind of stuff could be set in DSN only. Especially since predis seems to be of same opinion with their verify_peer option, see docs at  https://github.com/predis/predis and https://github.com/predis/predis/blob/b060d431574e2b637b266dc83d5d28dff03cd712/src/Connection/StreamConnection.php#L32

docs/README.md
```diff
@@ -293,6 +293,10 @@ snc_redis:
                 iterable_multibulk: false
                 throw_errors: true
                 cluster: predis
+                # Here you can specify additional context data, see connect/pconnect documentation here
+                # https://github.com/phpredis/phpredis#connect-open
+                # Stream configuration options can be found here https://www.php.net/manual/en/context.ssl.php
+                context: {'stream': {'verify_peer': false, 'allow_self_signed': true, 'verify_peer_name': false}}
```
ostrolucky, line 299: ssl_context?

ostrolucky (changes requested):
Well, we can have this as array if you really want, but we should have support for predis definitely too. And specifying this via DSNs.

src/DependencyInjection/Configuration/Configuration.php
```diff
@@ -119,6 +119,7 @@ private function addClientsSection(ArrayNodeDefinition $rootNode): void
                                     ->booleanNode('connection_async')->defaultFalse()->end()
                                     ->booleanNode('connection_persistent')->defaultFalse()->end()
                                     ->scalarNode('connection_timeout')->cannotBeEmpty()->defaultValue(5)->end()
+                                    ->variableNode('ssl_context')->defaultNull()->end()
```
ostrolucky, line 122: This should go under `parameters`, like all other connection related settings.

mattiabasone: @ostrolucky ok, I'll move this option under `paramenters`

.github/workflows/continuous-integration.yml
```diff
@@ -72,8 +72,10 @@ jobs:
           extensions: "${{ matrix.php-extensions }}"
           ini-values: "zend.assertions=1, max_execution_time=30"
 
-      - name: "Install symfony/flex"
-        run: "composer global require --no-progress --no-scripts --no-plugins symfony/flex"
+#      - name: "Install symfony/flex"
+#        run: |
+#          "composer global config --no-plugins allow-plugins.symfony/flex true"
+#          "composer global require --no-progress --no-scripts --no-interaction --no-plugins symfony/flex"
```
mattiabasone, line 78: @ostrolucky can we remove this CI step or is necessary? It is causing some problems ([error](https://github.com/snc/SncRedisBundle/runs/7279992495?check_suite_focus=true#step:11:23))

ostrolucky: It's necessary to make SYMFONY_REQUIRE env var working. We need to find out why this stopped working

mattiabasone: I've noticed that you have just fixed on branch master the composer error. 
FYI if you don't want to modify `composer.json` it works also with these [changes](https://github.com/snc/SncRedisBundle/pull/669/commits/214f7990f92c2f6d5cdf4b13eaad1213cc3f1cf9)

ostrolucky: I think it's better to modify composer.json because that's where we already allowed another plugin too. Can you rebase your branch on master?

mattiabasone: Done :)

PR #675, opened 2022-10-26
Fix ConnectionWrapper with aggregated connections

specdrum-agc:
Fixes https://github.com/predis/predis/issues/777

### DIFF
--- a/src/Client/Predis/Connection/ConnectionWrapper.php
+++ b/src/Client/Predis/Connection/ConnectionWrapper.php
@@ -15,6 +15,7 @@ namespace Snc\RedisBundle\Client\Predis\Connection;
 
 use Closure;
 use Predis\Command\CommandInterface;
+use Predis\Connection\ConnectionException;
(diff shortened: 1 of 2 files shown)


ostrolucky:
Sorry, I have tested this, but it doesn't solve described issue. Try this configuration
```yaml
snc_redis:
        predis:
            type: predis
            alias: predis
            dsn:
                - redis://127.0.0.1:9999
                - redis://127.0.0.1:6379
            options:
                cluster: redis
```

then execute any command via `bin/console redis:query --client=predis`. With this patch, you will still get
```
In ConnectionWrapper.php line 169:

  Connection refused [tcp://127.0.0.1:9999]
```

Once  the config is switched from predis to phpredis, it works correclty then and it fails over to next redis node.

specdrum-agc:
My bad, I passed wrong instance to the exception.

ostrolucky:
Great, this one is working, thx. I'll release now

PR #677, opened 2022-11-05
Reset the logger after the collector reset

tourze:
\Snc\RedisBundle\Logger\RedisLogger is not implement ResetInterface now, so in some special runtime (like swoole), RedisDataCollector will not destroy previous request's redis command logs. It is better to reset the logger at the time of the collector reset, then developer can decorate the \Snc\RedisBundle\Logger\RedisLogger and reset it.

### DIFF
--- a/src/DataCollector/RedisDataCollector.php
+++ b/src/DataCollector/RedisDataCollector.php
@@ -35,6 +35,9 @@ class RedisDataCollector extends DataCollector
     public function reset(): void
     {
         $this->data = [];
+        if ($this->logger instanceof ResetInterface) {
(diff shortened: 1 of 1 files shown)


ostrolucky:
Correct solution would be to make RedisLogger implement ResetInterface

tourze:
> Correct solution would be to make RedisLogger implement ResetInterface, wouldn't it?

You are right.

ostrolucky:
Fixed in https://github.com/snc/SncRedisBundle/commit/cbbaf9e2b82a221961025f8e1c0c27f2a7e32cfa

Issue #679, opened 2022-11-10
Supporting Relay


tillkruss:
Hey 👋

I'm the maintainer of Predis and teamed up with the maintainer of PhpRedis, @michael-grunder to build a new Redis extension for PHP called [Relay](https://github.com/cachewerk/relay) and we'd love to integrate with Symfony.

We're a bit clueless when it comes to Symfony and it's ecosystem and are looking for advice.

Symfony seems to have it's own Redis components, what's the difference to this repo?
Should we add support for Relay via this repo, or fork it?

ostrolucky:
What this bundle does is basically:
- provides a profiling/debugging panel, as well provides data for other panels
- integrates logging for the queries being made
- provides client independent configuration of redis client with validation, instead of getting errors during runtime

And yeah, Relay would be a welcome addition. Check PhpredisClientFactory for inspiration, you should be able to figure out what should be adapted for Relay. Rest of the work is not much.

That said, I think what this bundle needs more than Relay support at the moment is Predis 2.x support.

PR #688, opened 2022-12-19
Add Relay\Relay support

ostrolucky:
This adds support for new [relay](https://relay.so/) redis extension.

Closes https://github.com/snc/SncRedisBundle/issues/679

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,7 +2,7 @@
 
 ## About ##
 
-This bundle integrates [Predis](https://github.com/nrk/predis) and [phpredis](https://github.com/nicolasff/phpredis) into your Symfony application.
+This bundle integrates [Predis](https://github.com/nrk/predis), [PhpRedis](https://github.com/nicolasff/phpredis) and [Relay](https://relay.so/) into your Symfony application.
 
 ## Installation ##
 
@@ -52,7 +52,7 @@ You have to configure at least one client. In the above example your service
 container will contain the service `snc_redis.default` which will return a
 `Predis` client.
 
-Available types are `predis` and `phpredis`.
+Available types are `predis`, `phpredis` and `relay`.
 
 A more complex setup which contains a clustered client could look like this:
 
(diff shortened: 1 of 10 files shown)


docs/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## About ##
 
-This bundle integrates [Predis](https://github.com/nrk/predis) and [phpredis](https://github.com/nicolasff/phpredis) into your Symfony application.
+This bundle integrates [Predis](https://github.com/nrk/predis), [phpredis](https://github.com/nicolasff/phpredis) and [relay](https://relay.so/) into your Symfony application.
```
tillkruss, line 5: You could spell it PhpRedis and Relay.

src/Factory/PhpredisClientFactory.php
```diff
@@ -229,24 +242,18 @@ private function createClient(RedisDsn $dsn, string $class, string $alias, array
...
-            'default' => Redis::SERIALIZER_NONE,
-            'json' => Redis::SERIALIZER_JSON,
-            'none' => Redis::SERIALIZER_NONE,
-            'php' => Redis::SERIALIZER_PHP,
+            'default' => 0, // Redis::SERIALIZER_NONE,
+            'none' => 0, // Redis::SERIALIZER_NONE,
+            'php' => 1, //Redis::SERIALIZER_PHP,
+            'igbinary' => 2, //Redis::SERIALIZER_IGBINARY,
+            'msgpack' => 3, //Redis::SERIALIZER_MSGPACK,
+            'json' => 4, // Redis::SERIALIZER_JSON,
```
tillkruss, line 254: Those constants are identical between with Relay and PhpRedis if you want to keep using them.

ostrolucky: Yeah that's why I hardcoded the values. Because I want that piece of code working no matter if Phpredis is installed or not. Easier this way rather than adding conditional to know which class to use where to get constant values from.

Issue #692, opened 2023-01-06
replication option: predis/predis v2 doesn't allow `true`


Gregoire-M:
I'm updating a project that uses predis and SncRedisBundle with this config:

```
  - Upgrading predis/predis (v1.1.10 => v2.0.3)
  - Upgrading snc/redis-bundle (4.4.1 => 4.5.0)
```

```yaml
snc_redis:
    clients:
        default:
            type: predis
            alias: default
            dsn:
                - "%env(REDIS_URL)%?alias=master"
                - "%env(REDIS_URL_READ_ONLY)%"
            logging: '%kernel.debug%'
            options:
                replication: true

```

But I get this error when I try to connect to Redis:

```
In Replication.php line 40:

  [InvalidArgumentException]
  Predis\Configuration\Option\Replication expects either a string or a callable value, boolean given


Exception trace:
  at /srv/app/vendor/predis/predis/src/Configuration/Option/Replication.php:40
 Predis\Configuration\Option\Replication->filter() at /srv/app/vendor/predis/predis/src/Configuration/Options.php:103
 Predis\Configuration\Options->__get() at /srv/app/vendor/predis/predis/src/Client.php:125
 Predis\Client::createConnection() at /srv/app/vendor/predis/predis/src/Client.php:61
 ...
```

Indeed, in [predis code](https://github.com/predis/predis/blob/main/src/Configuration/Option/Replication.php#L31-L46), I can see that the value is supposed to be a string (`predis`, `sentinel` or `redis-sentinel`) or a callable.

But SncRedisBundle [forces](https://github.com/snc/SncRedisBundle/blob/master/src/DependencyInjection/Configuration/Configuration.php#L130) me to set `true` or `'sentinel'`.

Am I missing something? Or is there a problem between predis v2 and SncRedisBundle v4.5?

ostrolucky:
Yes. predis 2 support is quite new. Thanks for reporting

franmomu:
I'll try to take a look later today, I guess we should allow `predis` and `redis-sentinel`, treat `true` as `predis`, and deprecate `true`.

PR #693, opened 2023-01-06
Fix predis replication using true as config value for predis

franmomu:
Fixes https://github.com/snc/SncRedisBundle/issues/692

It also allows to use `predis` as replication value, I didn't add `redis-sentinel` as it works like `sentinel`. Since the `true` value is valid for phpredis and relay, we can leave it like this I guess.

### DIFF
--- a/src/DependencyInjection/Configuration/Configuration.php
+++ b/src/DependencyInjection/Configuration/Configuration.php
@@ -127,7 +127,7 @@ class Configuration implements ConfigurationInterface
                                     ->scalarNode('serialization')->defaultValue('default')->end()
                                     ->scalarNode('cluster')->defaultNull()->end()
                                     ->scalarNode('prefix')->defaultNull()->end()
-                                    ->enumNode('replication')->values([true, 'sentinel'])->end()
+                                    ->enumNode('replication')->values([true, 'predis', 'sentinel'])->end()
                                     ->scalarNode('service')->defaultNull()->end()
                                     ->enumNode('slave_failover')->values(['none', 'error', 'distribute', 'distribute_slaves'])->end()
                                     ->arrayNode('parameters')
--- a/src/DependencyInjection/Configuration/Configuration.php
+++ b/src/DependencyInjection/Configuration/Configuration.php
@@ -30,6 +30,7 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
 
 use function class_exists;
 use function is_iterable;
+use function trigger_deprecation;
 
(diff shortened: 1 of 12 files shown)


ostrolucky:
Can we have a test simulating scenario from https://github.com/snc/SncRedisBundle/issues/692?

franmomu:
From https://github.com/predis/predis/blob/main/CHANGELOG.md#v200-2022-06-08

> Replication backends now use the role parameter instead of alias in order to distinguish the role of a connection. 

There was also this change

ostrolucky:
Thx, I've adjusted some things:
- Moved the BC layer into Configuration instead of SncRedisExtension
- Added deprecation of `true` value
- Removed BC layer for changing `alias` to `role` if `alias` is master, I think that's too much to do on our side and might apply `master` role even if it was actually meant to be purely alias only

ostrolucky (approved):

franmomu:
> * Removed BC layer for changing `alias` to `role` if `alias` is master, I think that's too much to do on our side and might apply `master` role even if it was actually meant to be purely alias only

I think I would keep the BC layer for `alias` and `role`, otherwise using the configuration from https://github.com/snc/SncRedisBundle/issues/692 won't work, maybe only add that BC layer when the `replication` option is `predis`?

ostrolucky:
I know it won't work, but I would argue that's not a fault of bundle, but predis, similarly as we had to remove `profile` option. Predis used to have `alias` only, now it has `alias` and `role` and I don't think we should assume user means `role` when they configure `alias`.

src/DependencyInjection/Configuration/Configuration.php
```diff
@@ -127,7 +128,21 @@ private function addClientsSection(ArrayNodeDefinition $rootNode): void
...
                                     ->scalarNode('prefix')->defaultNull()->end()
-                                    ->enumNode('replication')->values([true, 'sentinel'])->end()
+                                    ->enumNode('replication')
+                                        ->values([true, 'predis', 'sentinel'])
+                                        ->beforeNormalization()
+                                            ->ifTrue(static fn ($v) => $v === true)
+                                            ->then(static function () {
+                                                trigger_deprecation(
+                                                    'snc/redis-bundle',
+                                                    '4.6',
```
OskarStark, line 138: 4.5.1 ?

ostrolucky: Ah yeah sorry my oversight, should have tagged 4.6. Anyways I'll leave this like this now because it's not a big issue

PR #694, opened 2023-01-06
Use R2 for Relay

tillkruss:
We're switching from S3 to R2.


szepeviktor:
@tillkruss You could redirect requests to S3 to R2 and keep the bucket for 1 more year.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-page-redirect.html

tillkruss:
@szepeviktor: Sadly won't work in our case.

ostrolucky (approved):

Issue #697, opened 2023-02-09
Phpredis sentinel auth


McFistyBuns:
Please correct me if I missed something, but it doesn't look like you can pass any password to a sentinel when using Phpredis.
https://github.com/snc/SncRedisBundle/blob/dc5f4438d669f1c52234f1d2b7bc558f8399c58e/src/Factory/PhpredisClientFactory.php#L124-L131
the Phpredis RedisSentinel class allows for an $auth parameter at the end.

https://github.com/phpredis/phpredis/blob/develop/sentinel.md#readme

Took me a bit to figure out why my setup was working with predis and not phpredis.  I patched in the parameter for myself to pass the password from the parameters config: `$options['parameters']['password'] ?? null` since my cluster setup uses the same password for both.  I could see a more robust path, though, in checking the dsn string for a password first in case the the sentinel and the node passwords are different.

ostrolucky:
That's correct. PR welcome ;)

ostrolucky:
Done in #727

PR #698, opened 2023-02-09
Added phpredis sentinel auth

McFistyBuns:
Updated the PhpredisClientFactory to allow for password authentication when using redis sentinel.  The default is to use the password set in the parameters password configuration, but also allows for an override in the dsn.

https://github.com/snc/SncRedisBundle/issues/697#issue-1578053433


ostrolucky:
I've worked on this, but I realized this approach cannot work as is, because it requires having same password on both: sentinel and redis servers. In case somebody had password on redis servers only and sentinel unprotected, this change will attempt to auth the sentinel now and connection will fail. Needs some rethinking.

ostrolucky:
Predis also doesn't support this https://github.com/predis/predis/issues/594 And they have a same concern as I have pointed out: username/password cannot be reused for sentinel, there should be different configuration. And at this point I don't really want to introduce it, because if we introduce new option and later predis introduces separate one, it will most likely be a different option and we would then have to support both :/

ostrolucky:
Superseded by https://github.com/snc/SncRedisBundle/pull/727

PR #707, opened 2023-07-06
Fix(Readme): fix config replication true

malekbig:
fix docs options replication equal true or false


ostrolucky:
Umm https://github.com/snc/SncRedisBundle/pull/693/files#diff-4e41bb7ece752048ea9a68d151a55920b7613e2e9430ef7714db42cf0933589cR139

PR #716, opened 2023-11-01
Fix installing symfony versions in CI

franmomu:
I've seen the new release and I was checking the versions installed and for example for:
- 4.4: https://github.com/snc/SncRedisBundle/actions/runs/6712763541/job/18243020072
- 5.4: https://github.com/snc/SncRedisBundle/actions/runs/6712763541/job/18243020270

It's installing the packages in the flex step and using incorrect versions.

I've removed `psalm/vimeo` in the job when the using Symfony 4.4 because [it requires `"symfony/filesystem": "^5.4 || ^6.0",`](https://github.com/vimeo/psalm/blob/4e177bf0c9f03c17d2fbfd83b7cc9c47605274d8/composer.json#L38).

I've added the "prefer-stable" section since it was installing dev versions for stable ones (4.4 and 5.4).

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,9 +47,12 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           extensions: "${{ matrix.php-extensions }},relay"
           ini-values: "zend.assertions=1, max_execution_time=30"
+          tools: "flex"
 
-      - name: "Install symfony/flex"
-        run: "composer require --no-progress --no-scripts --no-plugins symfony/flex"
+      - name: "Remove psalm for Symfony 4.4"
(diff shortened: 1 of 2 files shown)


ostrolucky:
This bundle doesn't use symfony/filesystem, so I would say its version doesn't matter. Fixing installation of other packages when installing flex is good that you fixed, though.

franmomu:
> This bundle doesn't use symfony/filesystem, so I would say its version doesn't matter. Fixing installation of other packages when installing flex is good that you fixed, though.

Apparently it does:

https://github.com/snc/SncRedisBundle/blob/ff9c4e6f8a8292f2549f07c5c303d97d8e19374d/composer.json#L44 😞

PR #722, opened 2024-02-12
[fix] Pass on username parameter from dsn to Predis Factory

rvdbogerd:
After debugging why I could use my application with default credentials, but not with a newly created acl user, I found that the username parameter is not passed in the PredisParametersFactory, when it is defined in the Dsn.

This PR fixes the issue.

### DIFF
--- a/src/Factory/PredisParametersFactory.php
+++ b/src/Factory/PredisParametersFactory.php
@@ -58,6 +58,7 @@ class PredisParametersFactory
         if ($dsn->getUsername() !== null) {
             $options['username'] = $dsn->getUsername();
         }
+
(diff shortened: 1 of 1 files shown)


src/Factory/PredisParametersFactory.php
```diff
@@ -55,6 +55,10 @@ private static function parseDsn(RedisDsn $dsn): array
             }
         }
 
+        if ($dsn->getUsername() !== null) {
+            $options['username'] = $dsn->getUsername();
+        }
+
```
rvdbogerd, line 61: ``
```suggestion

```

PR #723, opened 2024-03-08
Fix Relay\Sentinel compatibility

kkrasnik:


ostrolucky:
Looking good. Just fix cs.

PR #727, opened 2024-09-04
Add phpredis sentinel auth support

vigikaran:


ostrolucky:
This will break BC because people's sentinel instances will suddenly try to authenticate. Let's have only `sentinel_auth` option (null by default). If people want to re-use same password for sentinel as for the actual redis server instances, they can repeat the value.

You should also cover this with tests, for inspiration you can check https://github.com/snc/SncRedisBundle/pull/698

PR #728, opened 2024-09-12
Ignore trailing slash in DSN

HypeMC:
Currently the DSN doesn't work if you have an optional db, eg:

```yaml
snc_redis:
  clients:
    default:
      dsn: '%env(REDIS_DSN)%/%env(default::REDIS_DB)%'
```


ostrolucky:
I'm sorry but this would be quite inconsistent with other projects. AFAIK no project is doing this. If you have optional database, you can either include it in DSN, or combine dsn with specific parameters like so
```yaml
snc_redis:
  clients:
    default:
      dsn: '%env(REDIS_DSN)%'
      options:
        parameters:
          database: '%env(default::REDIS_DB)%'
```

HypeMC:
@ostrolucky Actually, it works with eg Symfony's lock component:

```yaml
framework:
  lock:
    resources:
      default: '%env(REDIS_DSN)%/%env(default::REDIS_DB)%'
```

In any case, there are some examples that need to be fixed then, eg https://github.com/snc/SncRedisBundle/tree/master/docs#usage-with-symfonyweb-profiler-bundle

Issue #735, opened 2025-05-06
Fatal error when using Predis v3.00


schodemeiss:
When upgrading to [Predis v3.0.0](https://github.com/predis/predis/releases/tag/v3.0.0) I get the following error in my Symfony application.

Fatal error: Class Snc\RedisBundle\Client\Predis\Connection\ConnectionWrapper contains 3 abstract methods and must therefore be declared abstract or implement the remaining methods (Predis\Connection\NodeConnectionInterface::getClientId, Predis\Connection\NodeConnectionInterface::write, Predis\Connection\NodeConnectionInterface::hasDataToRead) in /var/www/api/vendor/snc/redis-bundle/src/Client/Predis/Connection/ConnectionWrapper.php on line 34

I guess the paths to fixing this are for this bundle to know the difference between the versions and support them; or signal the incompatability when installing Predis v3.x via composer.json.

PR #736, opened 2025-05-15
chore(deps): support predis v3

p4veI:
Resolves: #735


composer.json
```diff
@@ -36,7 +36,7 @@
         "friendsofphp/proxy-manager-lts": "^1.0.6",
         "monolog/monolog": "*",
         "phpunit/phpunit": "^9.5.28 || ^10",
-        "predis/predis": "^2.0",
+        "predis/predis": "^3.0",
```
Seldaek, line 39:
```suggestion
        "predis/predis": "^2.0 || ^3.0",
```

That way the lowest-deps CI builds can ensure compatibility with 2.x still.

PR #737, opened 2025-05-26
[Dependency] Added conflict with predis >= 3.0

popy-dev:
As of today, redis-bundle doesn't implements the latest predis interface. The suggested update will prevent installing/updating this version.

This would fix issue #735 until #736 gets finalized/merged.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
     },
     "conflict": {
         "ext-redis": "<5.3.2",
-        "predis/predis": "<2.0"
+        "predis/predis": "<2.0 || >=3.0"
     },
     "extra": {
         "branch-alias": {


Seldaek:
It won't really fix anything because Composer will just use an older version of the bundle then that has no conflict. This should have been there from the start ideally, but still good to merge IMO for clarity.

Seldaek:
No what I mean is that if we merge/release this PR, a composer update would still then possibly upgrade to predis/predis 3.x if your requirements allow it, but it would keep this bundle at version 2.9.0 and not upgrade it to the 2.9.1 that has a conflict against predis 3.

ostrolucky:
not needed thanks to https://github.com/snc/SncRedisBundle/pull/736

PR #739, opened 2025-06-26
Add connection_persistent_id parameter

### LINKED ISSUE #738: Add a way to specify the conn_uid for predis persistent connections
See https://github.com/predis/predis?tab=readme-ov-file#persistent-connections - namely this:

> If you operate on multiple clients within the same application, and they communicate with the same resource, by default they will share the same socket (that's the default behaviour of persistent sockets). So in this case you would need to additionally provide a conn_uid identifier for each client, this way each client will create its own socket so the connection context won't be shared across clients.

I would like to pass that conn_uid but I can't seem to do that as I don't see a param like connection_persistent..

Seldaek:
Fixes #738


Seldaek:
I personally find it more difficult to work with because you have less clear types, makes the documentation, autocomplete, more difficult, it needs to be validated by hand then, etc.

Simpler types usually are better, but if you really want I can also change it I don't mind THAT much :D 

In the meantime I'll go fix the CS issues.

ostrolucky:
Right. But this ID makes more sense only for persistent connections. If persistent connection is disabled, then it introduces undefined state when ID is defined. User might then not be sure if it's going to error out, or persistent connection is disabled, or enabled. It might also confuse people familiar with phpredis, because there once you set the ID, it's going to enable persistent connections.  I don't mind THAT much either way, though either :D

ostrolucky:
Tagged in 4.10.0

Seldaek:
Thanks!

PR #744, opened 2025-07-16
refactor: add closure return type

shakaran:


shakaran:
> Please stop making code style contributions

I won't send additional contributions related to code style if you don't approve or you don't want it (second PR that it gets rejected), but could you elaborate on why these aren't suitable for the project or consider adding typed support which respects PHP 7.4?

I am trying to help with small contributions to learn and know better the project. Thanks for you time.

shakaran:
> y click of a button, since these changes can be done by code style fixers. That just further points out that time should be spent

I am using rector and custom tools for this. Would be appreciated if I send a PR with the rector config and rector/rector as dev dependency? So the project could run exact rules and it even could be added in the CI/CD and following the current doctrine/coding-standard also. Also with the rules skipped that you don't want it or you don't approved previously. 

Let me know if it is a good idea sent the PR with the rector rules.

ostrolucky:
I don't really want to introduce another such tool in the project, since we already use PHP_Codesniffer. These rules should be possible to set up there.

shakaran:
> I don't really want to introduce another such tool in the project, since we already use PHP_Codesniffer. These rules should be possible to set up there.

I understand the hesitation to add another tool, especially when the project already using PHP_CodeSniffer. However, Rector and PHPCS serve very different purposes:

PHP_CodeSniffer enforces coding standards and formatting rules. It mainly handles whitespace, naming conventions, and style-related issues. It doesn’t change the underlying logic of the code.

Rector, on the other hand, performs automatic refactoring and code transformations. It can upgrade syntax, replace deprecated APIs, and modernize codebases (for example, migrating old PHP patterns to newer ones). These kinds of changes are not something PHP_CodeSniffer can do.

Without Rector, these kinds of updates would have to be manually written and reviewed, which is time-consuming and error-prone. So I think that it is more like complementary, not redundant since Rector doesn't replace PHP_CodeSniffer

PR #745, opened 2025-07-16
feat: phpredis: add support for OPT_SCAN 

### LINKED ISSUE #608: Allow setting the OPT_SCAN option for phpredis
I need to be able to do:

```
$redis->setOption(Redis::OPT_SCAN, Redis::SCAN_PREFIX);
```

But right now `PhpredisClientFactory` doesn't allow to do it.

shakaran:
Fixes #608


src/Factory/PredisParametersFactory.php
```diff
@@ -24,8 +25,13 @@ public static function create(array $options, string $class, string $dsn): Param
         }
 
         $defaultOptions = ['timeout' => null]; // Allow to be consistent with old version of Predis where default timeout was 5
-        $dsnOptions     = static::parseDsn(new RedisDsn($dsn));
-        $dsnOptions     = array_merge($defaultOptions, $options, $dsnOptions);
+
+        if (isset($options['scan']) && $options['scan'] === 'prefix') {
+            $dsnOptions[Redis::OPT_SCAN] = Redis::SCAN_PREFIX;
```
ostrolucky, line 30: This is overwritten by next instruction

ostrolucky, line 30: This class shouldn't use \Redis class, as that belongs to phpredis

ostrolucky (changes requested):

src/Factory/PredisParametersFactory.php
shakaran: Resolved at https://github.com/snc/SncRedisBundle/pull/745/commits/95f7e4e6cc6ba8c980358c90dc33a809842bcca2 using class constants

shakaran: You are right, moved after parseDsn() so the options get merged. Resolved at https://github.com/snc/SncRedisBundle/pull/745/commits/95f7e4e6cc6ba8c980358c90dc33a809842bcca2

ostrolucky:
Sorry, in the end found this option is not relevant for predis so I've removed it. What was mistaking me was that there was never added support for `prefix` in this bundle for predis. In master I've added support for `prefix` in Predis. I've rebased this PR  and did bunch of changes, most of it was a cleanup but one thing that's functional is that config option is now numeric, this way people can directly reference values of the phpredis constants.

Issue #195, opened 2015-08-27
RedisArray on phpredis


AlmogBaku:
Is there any progress about this?

snc:
What do you mean exactly?

AlmogBaku:
Currently- the bundle doesn't support redis-array functionality with the native phpredis but only with predis.
But as far as I know, this feature is feature should be exists: https://github.com/phpredis/phpredis/blob/master/arrays.markdown#readme

AlmogBaku:
https://github.com/snc/SncRedisBundle/blob/23ffb61ad14aeb5086f053731b8476b54a5b6fd2/DependencyInjection/SncRedisExtension.php#L224

dklesev:
any update on this?

ste93cry:
Is `RedisArray` unsupported for some specific reason or is it just because it didn't get implemented? In the latter case I could take a look and contribute the change

ostrolucky:
Go ahead, we are waiting for somebody contribute this

guillaumedelre:
PR #758 adds support for `RedisArray` via a new `array: true` option under `clients.options` (phpredis only).

PR #746, opened 2025-07-16
feat: implement redis array on phpredis. Issue #195

shakaran:
Related #195


shakaran:
@ostrolucky Could you give some guide here? I don't know exactly which hosts or ports should use for the tests and get the connection done. I try to use like in other tests but it seems not working in the connection pipeline. Thanks

ostrolucky:
Well you are using made up API RedisArray::getHosts(), that's why.

PR #747, opened 2025-07-17
feat(ci): add composer cache in github ci

shakaran:


ostrolucky:
Thank you for thinking of this, however ramsey/composer-install that we use does this already

PR #748, opened 2025-07-17
feat(composer): add validate strict changes for composer.json

shakaran:
This PR allows to run successfully 

`composer validate --strict`

And avoid all the warnings.

The important changes:

- Minimum-stability should not be in a bundle:
Bundles and libraries should be agnostic about minimum stability, as they are dependencies of other projects. Defining minimum-stability in a bundle could force consuming projects to adopt an undesired stability level, which may lead to conflicts or instability.

Composer's strict mode only allows this in projects with "type": "project", and not for "type": "symfony-bundle".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,6 @@
     ],
     "homepage": "https://github.com/snc/SncRedisBundle",
     "license": "MIT",
-    "minimum-stability": "dev",
     "authors": [
         {
             "name": "Henrik Westphal",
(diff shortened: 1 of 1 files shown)


ostrolucky:
Replaced with https://github.com/snc/SncRedisBundle/commit/3829ae03f0b523f9a9e11215164f83ed7d611a47

Not interested in other changes. minimum-stability doesn't affect consumers, it affects people developing this bundle and running composer install locally. In that case I want to pull dev dependencies to ensure bundle works with those. I don't need passing strict mode of composer validate either.

PR #759, opened 2026-05-11
feat(predis): support json:/csv: env processors for DSN arrays

### LINKED ISSUE #483: SncRedis support for json environment variable type
SncRedis config does not support env variables of "json" type, at least for `dsn` property.
Example .env
```
REDIS_SENTINELS="[\"redis://redis-sentinel1.dev:26379\", \"redis://redis-sentinel2.dev:26379\", \"redis://redis-sentinel3.dev:26379\"]"
```
Example config
```
snc_redis:
    clients:
        default:
            type: predis
            alias: default
            dsn: '%env(json:REDIS_SENTINELS)%'
```

PHP warnings 
```
PHP Warning:  strrpos() expects parameter 1 to be string, array given in /home/app/src/vendor/snc/redis-bundle/DependencyInjection/Configuration/RedisDsn.php on line 172
PHP Warning:  preg_match() expects parameter 2 to be string, array given in /home/app/src/vendor/snc/redis-bundle/DependencyInjection/Configuration/RedisDsn.php on line 185
PHP Warning:  preg_match() expects parameter 2 to be string, array given in /home/app/src/vendor/snc/redis-bundle/DependencyInjection/Configuration/RedisDsn.php on line 190
PHP Warning:  preg_match() expects parameter 2 to be string, array given in /home/app/src/vendor/snc/redis-bundle/DependencyInjection/Configuration/RedisDsn.php on line 203
PHP Warning:  strpos() expects parameter 1 to be string, array given in /home/app/src/vendor/snc/redis-bundle/DependencyInjection/Configuration/RedisDsn.php on line 212
```

JSON support would simplify SncRedis config as sentinel dsn count may differ in different enviroments.

guillaumedelre:
## Summary

- Add `PredisParametersFactory::createFromDsns()` accepting `string|array` to handle DSNs resolved at runtime by Symfony env processors
- `SncRedisExtension` now uses `createFromDsns` (instead of `create`) when the DSN is a `RedisEnvDsn`, so that `%env(json:REDIS_DSNS)%` or `%env(csv:REDIS_DSNS)%` resolve correctly at runtime
- Handles nested-array edge case (`[['redis://host1', 'redis://host2']]`) produced when a `json:` processor wraps a `csv:` result

## Motivation

Fixes #483. Using `%env(json:REDIS_DSNS)%` as a predis DSN caused a `TypeError` at runtime because `PredisParametersFactory::create()` expects `string $dsn` but Symfony's `json:` env processor resolves to an `array`.

## Test plan

- [x] `PredisParametersFactoryTest`: 4 new cases covering `createFromDsns` with string, single-element array, multiple DSNs, and nested array
- [x] `SncRedisExtensionEnvTest`: new `testPredisJsonEnvDsn` fixture + updated factory assertions for existing env tests
- [x] PHPCS clean, Psalm 0 errors on changed files


psalm.xml.dist
```diff
@@ -19,6 +19,11 @@
                 <referencedMethod name="Symfony\Component\Config\Definition\Builder\NodeBuilder::end"/>
             </errorLevel>
         </UnusedMethodCall>
+        <PossiblyUnusedMethod>
+            <errorLevel type="suppress">
+                <referencedMethod name="Snc\RedisBundle\Tests\Functional\App\Controller\Controller::__construct"/>
+            </errorLevel>
+        </PossiblyUnusedMethod>
```
ostrolucky: Why?

tests/Functional/App/config.yaml
```diff
@@ -35,11 +35,9 @@ snc_redis:
         cluster:
             type: predis
             alias: cluster
-            dsn:
-              - redis://sncredis@127.0.0.1/3
-              - redis://sncredis@127.0.0.1/4
-              - redis://sncredis@127.0.0.1/5
+            dsn: "%env(json:REDIS_DSNS)%"
             options:
+                cluster: predis
```
ostrolucky: why?

tests/Functional/App/Controller/Controller.php
```diff
@@ -13,18 +13,25 @@
...
 use Symfony\Component\HttpFoundation\JsonResponse;
 
 class Controller extends AbstractController
 {
-    public function __invoke(Redis $redis): JsonResponse
+    public function __construct(
+        #[Autowire(service: 'snc_redis.cluster')]
+        private ClientInterface $cluster,
+    ) {
+    }
```
ostrolucky: let's inject all clients, iterate over them and make existing calls on all of them

tests/Factory/PredisParametersFactoryTest.php
```diff
@@ -165,4 +165,40 @@ public function testCreateException(): void
...
+    public function testCreateFromDsnsNestedArray(): void
+    {
+        $result = PredisParametersFactory::createFromDsns([], Parameters::class, [['redis://host1', 'redis://host2']]);
+        $this->assertIsArray($result);
+        $this->assertCount(2, $result);
+        $this->assertInstanceOf(Parameters::class, $result[0]);
+        $this->assertInstanceOf(Parameters::class, $result[1]);
+        $this->assertSame('host1', $result[0]->toArray()['host']);
+        $this->assertSame('host2', $result[1]->toArray()['host']);
+    }
```
ostrolucky: useless and misleading, remove

src/Factory/PredisParametersFactory.php
```diff
@@ -9,13 +9,43 @@
...
 class PredisParametersFactory
 {
+    /**
+     * @param class-string<ParametersInterface>      $class
+     * @param array<string, mixed>                   $options
+     * @param string|list<string>|list<list<string>> $dsns
+     *
+     * @return ParametersInterface|list<ParametersInterface>
+     */
+    public static function createFromDsns(array $options, string $class, string|array $dsns): ParametersInterface|array
```
ostrolucky: modify existing factory instead, otherwise it's no longer publicly used

src/DependencyInjection/SncRedisExtension.php
```diff
@@ -184,7 +184,11 @@ private function loadPredisClient(array $client, ContainerBuilder $container): v
         $container->setDefinition($optionId, $optionDef);
         $clientDef = new Definition((string) $container->getParameter('snc_redis.client.class'));
         $clientDef->addTag('snc_redis.client', ['alias' => $client['alias']]);
-        if ($connectionCount === 1 && !isset($client['options']['cluster']) && !isset($client['options']['replication'])) {
+        // When a single env-var DSN uses json:/csv: processors it may resolve to N parameters at runtime;
+        // pass the Reference directly so Predis receives [p1,p2,...] instead of [[p1,p2,...]].
+        $singleEnvDsn = $connectionCount === 1 && $client['dsns'][0] instanceof RedisEnvDsn;
```
ostrolucky: better to fix $connectionCount and not do $wrongConnectionCount === 1 || $loophole

src/DependencyInjection/SncRedisExtension.php
```diff
@@ -184,7 +184,11 @@ private function loadPredisClient(array $client, ContainerBuilder $container): v
         $container->setDefinition($optionId, $optionDef);
         $clientDef = new Definition((string) $container->getParameter('snc_redis.client.class'));
         $clientDef->addTag('snc_redis.client', ['alias' => $client['alias']]);
-        if ($connectionCount === 1 && !isset($client['options']['cluster']) && !isset($client['options']['replication'])) {
+        // When a single env-var DSN uses json:/csv: processors it may resolve to N parameters at runtime;
+        // pass the Reference directly so Predis receives [p1,p2,...] instead of [[p1,p2,...]].
```
ostrolucky: what reference is being passed directly? comment doesn't make sense to me

tests/DependencyInjection/SncRedisExtensionEnvTest.php
```diff
@@ -36,11 +36,45 @@ public function testPredisDefaultParameterWithSSLContextConfigLoad(): void
...
+        );
+
+        $clientDefinition = $container->findDefinition('snc_redis.default');
+
+        // Single Reference (not [Reference]) prevents nested-array [[p1,p2]] when env resolves to multiple DSNs.
+        $this->assertSame(
+            'snc_redis.connection.default_parameters.default',
+            (string) $clientDefinition->getArgument(0),
+        );
+    }
```
ostrolucky: What are these testing? Even comment is misleading as I see no relation between it and assertion.

ostrolucky (changes requested):

tests/Functional/App/Controller/Controller.php
guillaumedelre: The intent is clear, but there are a few constraints worth clarifying before implementing:

The 6 configured clients split across two incompatible types (`Redis` for phpredis and `Predis\ClientInterface` for Predis), so a clean typed loop isn't straightforward. On top of that, `sentinel` and `with_acl` require special server setup not available in the test environment.

Three options we see:

1. **All Predis clients only** (`cache` + `cluster`) - both implement `ClientInterface`, clean typed loop, but leaves phpredis untested on this route.
2. **`default` (phpredis) + `cluster` (Predis)** - covers both driver types, requires `mixed` or `object` in the loop with a psalm suppress since they share no common interface.
3. **`TaggedIterator` on `snc_redis.client`** - auto-discovers all registered clients, same typing issue, but more future-proof.

Which direction would you prefer?

psalm.xml.dist
guillaumedelre: Psalm does not understand Symfony's `#[Autowire]` attribute for DI injection and incorrectly reports the constructor as possibly unused. The suppression avoids a false positive.

tests/Functional/App/config.yaml
guillaumedelre: The `cluster: predis` option is required here because the DSN is an env var resolved at runtime via `json:`. At compile time Symfony sees a single opaque placeholder, so it cannot count the connections — Predis needs to know the aggregation mode upfront in order to build the right connection factory.

tests/Functional/App/Controller/Controller.php
guillaumedelre: Note: the command count assertion in `IntegrationTest` will also need updating once this is settled, as it depends directly on how many clients are exercised in the controller.

psalm.xml.dist
ostrolucky: Add `<pluginClass class="Psalm\SymfonyPsalmPlugin\Plugin"/>` instead, that fixes it.

tests/Functional/App/Controller/Controller.php
ostrolucky: Option 3. I don't see a typing issue, they all have `get` and `set` and type can be typed as `iterable<ClientInterface|RedisArray>` etc.

Rebase the PR as well.

tests/Functional/App/config.yaml
ostrolucky: That's the whole point why env vars were not supported for Predis. Predis needs to know connectionCount at build time, where env vars are not resolved yet. Think of a solution.

tests/DependencyInjection/SncRedisExtensionEnvTest.php
ostrolucky: Remove these, it's better to have no tests than confusing tests.

tests/Factory/PredisParametersFactoryTest.php
```diff
@@ -165,4 +165,29 @@ public function testCreateException(): void
...
+        $this->assertInstanceOf(Parameters::class, $parameters);
+        $this->assertSame('localhost', $parameters->toArray()['host']);
+    }
+
+    public function testCreateWithSingleElementArray(): void
+    {
+        $parameters = PredisParametersFactory::create([], Parameters::class, ['redis://localhost']);
+        $this->assertInstanceOf(Parameters::class, $parameters);
+        $this->assertSame('localhost', $parameters->toArray()['host']);
+    }
```
ostrolucky: `@testWith`

tests/Functional/App/Controller/Controller.php
```diff
@@ -13,18 +13,26 @@
...
+    public function __construct(private iterable $clients)
     {
-        $redis->set('foo', 'bar');
+    }
+
+    public function __invoke(): JsonResponse
+    {
+        $result = null;
+        foreach ($this->clients as $client) {
+            /** @psalm-suppress MixedMethodCall */
```
ostrolucky: No suppressing, fix `@param` like mentioned in my comment

.github/workflows/redis-configs/gen-tls-certs.sh
```diff
@@ -12,4 +12,4 @@ mkdir -p /tmp/redis-tls
             2>/dev/null
         cp /tmp/redis-tls/server.crt /tmp/redis-tls/ca.crt
     fi
-) 9>/tmp/redis-tls.lock
```
ostrolucky, line 15: revert

ostrolucky (changes requested):

psalm.xml.dist
guillaumedelre: Regarding the `psalm/plugin-symfony` suggestion: I investigated adding it, but the plugin is not compatible with the current Psalm version range (`^5 || ^6 || ^7`) — `psalm/plugin-symfony` requires Psalm `^5` only and does not support `^6`/`^7`. Removing the suppression without the plugin causes CI to fail with `PossiblyUnusedMethod` on `Controller::__construct` since Psalm has no way to know Symfony's DI container calls it. I've kept the suppression as the least invasive option. Happy to revisit if there's a path forward here.

tests/Functional/App/Controller/Controller.php
```diff
@@ -13,18 +13,26 @@
...
 
+use Predis\ClientInterface;
 use Redis;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\JsonResponse;
 
 class Controller extends AbstractController
 {
-    public function __invoke(Redis $redis): JsonResponse
+    /** @param iterable<Redis|ClientInterface> $clients */
```
ostrolucky:
```suggestion
    /** @param iterable<Redis|ClientInterface|Relay|RedisArray> $clients */
```

psalm.xml.dist
ostrolucky: latest psalm/plugin-symfony supports ^6 and dev-master, which is ^7

ostrolucky (changes requested):

psalm.xml.dist
```diff
@@ -13,6 +14,18 @@
...
+                <referencedMethod name="Symfony\Component\Config\Definition\Builder\NodeBuilder::end"/>
+            </errorLevel>
+        </UnusedMethodCall>
+        <UndefinedInterfaceMethod>
+            <errorLevel type="suppress">
+                <referencedMethod name="Symfony\Component\Config\Definition\Builder\NodeParentInterface::end"/>
+            </errorLevel>
+        </UndefinedInterfaceMethod>
+    </issueHandlers>
     <plugins>
```
ostrolucky: better to do inline suppression

psalm.xml.dist
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <psalm
     errorLevel="4"
+    findUnusedIssueHandlerSuppression="false"
```
ostrolucky: this is a no go. perhaps can be done inline if possible

ostrolucky (changes requested):

PR #761, opened 2026-05-11
feat(dsn): add tls_version query parameter to force TLS version

### LINKED ISSUE #658: Add options to force TLS Version in the connection string

guillaumedelre:
## Summary

- Add a `tls_version` query parameter to the DSN to force a specific TLS protocol version
- For `phpredis`: changes the PHP stream wrapper prefix from `tls://` to `tlsv{version}://`
- For `predis`: injects `crypto_type` into the ssl connection options
- When both `tls_version` (DSN) and `ssl_context` (config) provide ssl options, they are merged

## Motivation

Closes #658 — PHP >= 7.4 defaults to TLS 1.3, which can cause random frozen connections with some Redis servers ([phpredis/phpredis#1726](https://github.com/phpredis/phpredis/issues/1726)). This implements the DSN-based approach discussed in the issue.

```yaml
snc_redis:
    clients:
        default:
            type: phpredis   # or predis
            dsn: 'rediss://my-redis-server:6379?tls_version=1.2'
```

Supported values: `1.0`, `1.1`, `1.2`. Compatible with Symfony environment variables:

```yaml
dsn: '%env(REDIS_URL)%'
# REDIS_URL=rediss://my-redis-server:6379?tls_version=1.2

dsn: 'rediss://my-redis-server:6379?tls_version=%env(REDIS_TLS_VERSION)%'
```

## Test plan

- [x] `testTlsVersion` — data provider covering all supported versions and the no-version case
- [x] Full `PredisParametersFactory` test suite: 163 tests, no regressions
- [x] PHPCS clean (`use const` for `STREAM_CRYPTO_METHOD_TLSv1_*_CLIENT`)
- [x] Psalm: no new errors vs baseline

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -356,6 +356,33 @@ snc_redis:
         key: monolog
 ```
 
+## Forcing the TLS version
+
+By default, PHP >= 7.4 negotiates TLS 1.3, which can cause random frozen connections with some Redis servers
+(see [phpredis#1726](https://github.com/phpredis/phpredis/issues/1726)).
+You can force a specific TLS version via the `tls_version` query parameter in the DSN:
+
+``` yaml
+snc_redis:
+    clients:
+        default:
+            type: phpredis   # or predis
+            dsn: 'rediss://my-redis-server:6379?tls_version=1.2'
+```
+
+Supported values: `1.0`, `1.1`, `1.2`.
+
+Works with Symfony environment variables:
+
+``` yaml
+# Whole DSN as env var (tls_version included)
+dsn: '%env(REDIS_URL)%'
+# REDIS_URL=rediss://my-redis-server:6379?tls_version=1.2
+
+# Only the version as env var
+dsn: 'rediss://my-redis-server:6379?tls_version=%env(REDIS_TLS_VERSION)%'
+```
+
 ## Usage with `symfony/web-profiler-bundle`
 
 If you are using [`symfony/web-profiler-bundle`](https://github.com/symfony/web-profiler-bundle)
--- a/docs/README.md
+++ b/docs/README.md
@@ -321,7 +321,7 @@ snc_redis:
         default:
             type: predis
             alias: default
-            dsn: redis://localhost
+            dsn: redis://localhost  # or: rediss://localhost?tls_version=1.2 to force TLS 1.2 (1.0, 1.1, 1.2 supported)
             logging: '%kernel.debug%'
             class: App\Redis\MyPredisClient
         cache:
@@ -354,33 +354,6 @@ snc_redis:
     monolog:
         client: cache
         key: monolog
-```
-
-## Forcing the TLS version
-
-By default, PHP >= 7.4 negotiates TLS 1.3, which can cause random frozen connections with some Redis servers
-(see [phpredis#1726](https://github.com/phpredis/phpredis/issues/1726)).
-You can force a specific TLS version via the `tls_version` query parameter in the DSN:
-
-``` yaml
-snc_redis:
-    clients:
-        default:
-            type: phpredis   # or predis
-            dsn: 'rediss://my-redis-server:6379?tls_version=1.2'
-```
-
-Supported values: `1.0`, `1.1`, `1.2`.
-
-Works with Symfony environment variables:
-
-``` yaml
-# Whole DSN as env var (tls_version included)
-dsn: '%env(REDIS_URL)%'
-# REDIS_URL=rediss://my-redis-server:6379?tls_version=1.2
-
-# Only the version as env var
-dsn: 'rediss://my-redis-server:6379?tls_version=%env(REDIS_TLS_VERSION)%'
 ```
 
 ## Usage with `symfony/web-profiler-bundle`
--- a/docs/README.md
+++ b/docs/README.md
@@ -321,9 +321,14 @@ snc_redis:
         default:
             type: predis
             alias: default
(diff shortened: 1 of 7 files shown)


src/Factory/PredisParametersFactory.php
```diff
@@ -28,7 +32,13 @@ public static function create(array $options, string $class, string $dsn): Param
 
         $defaultOptions = ['timeout' => null]; // Allow to be consistent with old version of Predis where default timeout was 5
         $dsnOptions     = static::parseDsn(new RedisDsn($dsn));
-        $dsnOptions     = array_merge($defaultOptions, $options, $dsnOptions);
+
+        // Merge ssl arrays so that DSN tls_version does not erase ssl_context from config
+        $sslOptions = array_merge($options['ssl'] ?? [], $dsnOptions['ssl'] ?? []);
+        $dsnOptions = array_merge($defaultOptions, $options, $dsnOptions);
+        if ($sslOptions !== []) {
+            $dsnOptions['ssl'] = $sslOptions;
```
ostrolucky: why override instead of merge?

docs/README.md
```diff
@@ -337,6 +337,33 @@ snc_redis:
...
+
+``` yaml
+# Whole DSN as env var (tls_version included)
+dsn: '%env(REDIS_URL)%'
+# REDIS_URL=rediss://my-redis-server:6379?tls_version=1.2
+
+# Only the version as env var
+dsn: 'rediss://my-redis-server:6379?tls_version=%env(REDIS_TLS_VERSION)%'
+```
+
```
ostrolucky: This new option is not worth adding such big chunk into docs. Especially since it seems to have been fixed upstream nowadays. Let's just add it to `Complete configuration example`

src/Factory/PredisParametersFactory.php
```diff
@@ -54,6 +64,11 @@ private static function parseDsn(RedisDsn $dsn): array
             $options['scheme'] = $dsn->getTls() ? 'tls' : 'tcp';
             $options['host']   = $dsn->getHost();
             $options['port']   = $dsn->getPort();
+
+            if ($dsn->getTls() && $dsn->getTlsVersion() !== null) {
+                $options['ssl'] = ['crypto_type' => self::tlsVersionToCryptoType($dsn->getTlsVersion())];
```
ostrolucky, line 68: use dynamic constant value retrieval via `constant()` value. That way we don't have keep appending to the list infinitely. I think implementation will be like so `constant('STREAM_CRYPTO_METHOD_TLSv'.str_replace('.', '_', '$dsn->getTlsVersion()).'_CLIENT')`

tests/DependencyInjection/Configuration/RedisDsnTest.php
```diff
@@ -301,4 +301,24 @@ public function testParameterValues(string $dsn, ?int $weight, ?string $alias):
         $this->assertSame($weight, $dsn->getWeight());
         $this->assertSame($alias, $dsn->getAlias());
     }
+
+    /** @return list<array{0: string, 1: ?string}> */
+    public static function tlsVersionValues(): array
```
ostrolucky, line 306: Also add tests to PhpredisClientFactoryTest and PredisParametersFactoryTest. Those are more useful.

ostrolucky (changes requested):

src/Factory/PredisParametersFactory.php
guillaumedelre: you're wright, let's merge :)

docs/README.md
```diff
@@ -303,7 +303,7 @@ snc_redis:
         default:
             type: predis
             alias: default
-            dsn: redis://localhost
+            dsn: redis://localhost  # or: rediss://localhost?tls_version=1.2 to force TLS 1.2 (1.0, 1.1, 1.2 supported)
```
ostrolucky: Ok I don't like this, let's add SSL example as another connection (perhaps named `secure`?). And don't specify which TLS versions are supported.

ostrolucky (changes requested):

PR #762, opened 2026-06-05
feat(phpredis): add compression and compression_level options

### LINKED ISSUE #331: Feature Request: Compression
I have been pondering the best place to request this, threads on redis discussions suggest within the app code to be the best place.

Compression of values can save at least 50% RAM in my Symfony app tests. Is this something SncRedisBundle can handle especially when Redis is used as a cache?

A see a few potential issues, though like if many App servers are connecting to a central Redis then they all must have the same compression library and version installed.

Thoughts?

koekaverna:
## Summary

- Adds `compression` config option mapped to `Redis::OPT_COMPRESSION` (7), supporting `none`, `lzf`, `zstd`, `lz4`
- Adds `compression_level` config option mapped to `Redis::OPT_COMPRESSION_LEVEL` (9)
- Applied to all phpredis client types: standard, cluster, and array

Closes #331

## Test plan

- [ ] `testLoadCompressionType` — verifies string→constant mapping per client factory
- [ ] `testLoadCompressionTypeFail` — verifies invalid type throws `InvalidConfigurationException`
- [ ] `testClientCompressionOption` — verifies config node is parsed and passed through correctly


src/Factory/PhpredisClientFactory.php
```diff
@@ -387,6 +411,23 @@ private function loadSerializationType(string $type): int
...
+        $types = [
+            'none' => 0, // Redis::COMPRESSION_NONE
+            'lzf'  => 1, // Redis::COMPRESSION_LZF
+            'zstd' => 2, // Redis::COMPRESSION_ZSTD
+            'lz4'  => 3, // Redis::COMPRESSION_LZ4
+        ];
+
+        if (array_key_exists($type, $types)) {
+            return $types[$type];
+        }
```
ostrolucky: Let's use `constant(Redis::COMPRESSION_.$type)` to avoid having to keep values in sync with constant

koekaverna: Done

